PR #28, opened 2023-12-23
Add PCS types and basic setup flow using SRS

dorebell:
This adds some basic support for generating public parameters from a SRS for a polynomial commitment scheme.


.gitignore
```diff
@@ -4,3 +4,6 @@ target
 .*.swp
 
 .config.env
+
+# public parameter files
+*.zst
```
slumber, line 9: This is supposed to used cache folder. If postponed for later, newline missing

prover/src/main.rs
```diff
@@ -38,6 +38,14 @@ enum Command {
...
         )]
         pp_file: String,
+
+        /// whether public parameters are compatible with proof compression
+        #[arg(short = 'c', long = "compressible", default_value = "false")]
+        com: bool,
+
+        /// SRS file: required with `--compressible`, otherwise ignored
+        #[arg(short = 's', long = "srs")]
+        srs_file: Option<String>,
```
slumber, line 48: Why are both needed if `srs_file: None` implies `com: false`? Or you can mark it with `require` from clap docs.

prover/src/pp.rs
```diff
@@ -28,28 +31,31 @@ where
...
 where
+    C: CommitmentScheme<P1>,
     SC: StepCircuit<F1> + Sync,
-    SP: SetupParams<G1, G2, C1, C2, RO, SC> + Sync,
+    SP: SetupParams<G1, G2, C, C2, RO, SC> + Sync,
 {
     let f = File::open(file)?;
     let mut dec = Decoder::new(&f)?;
-    let pp = PP::<SP>::deserialize_compressed(&mut dec)?;
+    let pp = PP::<C, SP>::deserialize_compressed_unchecked(&mut dec)?;
```
slumber, line 42: Why is this change needed?

prover/src/pp.rs
```diff
@@ -28,28 +31,31 @@ where
     Ok(())
 }
 
-pub fn load_pp<SP>(file: &str) -> Result<PP<SP>, ProofError>
+pub fn load_pp<C, SP>(file: &str) -> Result<PP<C, SP>, ProofError>
 where
+    C: CommitmentScheme<P1>,
     SC: StepCircuit<F1> + Sync,
-    SP: SetupParams<G1, G2, C1, C2, RO, SC> + Sync,
+    SP: SetupParams<G1, G2, C, C2, RO, SC> + Sync,
```
slumber, line 38: We should keep `C1 C2` notation

prover/src/pp.rs
```diff
@@ -76,23 +88,39 @@ pub fn gen_to_file(k: usize, par: bool, pp_file: &str) -> Result<(), ProofError>
     let _guard = term_ctx.display_step();
 
     if par {
-        let pp: ParPP = gen_vm_pp(k)?;
-        show_pp(&pp);
-        save_pp(pp, pp_file)
+        if com {
+            let srs_file = srs_file.ok_or(ProofError::MissingSRS)?;
+            println!("Loading SRS from {srs_file}...");
```
slumber, line 93: `nexus-tui` reserves the terminal window as a canvas and renders on it every 100ms. You will not see these prints.

prover/src/pp.rs
```diff
@@ -121,3 +149,28 @@ where
...
+        gen_test_srs_to_file(10, "small_test_srs.zst").unwrap();
+        let _srs: SRS = load_srs("small_test_srs.zst").unwrap();
+    }
+
+    #[test]
+    fn test_load_srs() {
+        test_srs_gen();
+        let srs: SRS = load_srs("small_test_srs.zst").unwrap();
+        println!("Loaded SRS for {} variables", srs.max_num_vars);
+    }
```
slumber, line 175: tests should not leave any files laying around

prover/src/pp.rs
dorebell: C1 is a type alias for Pedersen<G1>: here, we want C to be generic so it works
for PVC as well.

prover/src/pp.rs
dorebell: I thought it wouldn't make a difference anymore due to the arkworks update: is
this not true?

prover/src/pp.rs
dorebell: is there a way to delete the file here? we can just '#[ignore]' this test.

slumber: You shouldn't have any leftover even after running `cargo test --ignored`
One way is to make function generic over something that's `: Read` and test it with `& [u8]`, something of a mock interface.

As an example, [tools tests](https://github.com/nexus-xyz/nexus-zkvm/blob/d07d60698273a86be81b6ab76e66119b9d5347c1/tools/tools-dev/src/command/dev/config.rs#L143-L173) just use strings.

prover/src/pp.rs
slumber: Didn't notice it because of wildcard imports unfortunately

prover/src/srs.rs
```diff
@@ -0,0 +1,44 @@
...
+        srs.serialize_compressed(&mut enc)?;
+        enc.finish()?;
+        f.sync_all()?;
+        Ok(())
+    }
+
+    pub fn gen_test_srs_to_file(poly_length: usize, file: &str) -> Result<(), ProofError> {
+        let srs = gen_test_srs(poly_length)?;
+        save_srs(srs, file)
+    }
```
slumber, line 43: same story

prover/src/pp.rs
dorebell: hm, i'm not really sure how to change the functions here to work that way, since the files are going to be very big. I'm just going to take these tests out for now, since the compression-cli PR will let us test this functionality from the main interface.

slumber (approved):

PR #78, opened 2024-02-21
Avoids copying public parameters

dorebell:
This is an attempt to avoid making copies of portions of the SRS when calling the Zeromorph `trim()` function, by replacing vectors of group elements with CoW references (a similar technique is used in the `ark-poly-commit` crate). 

The compression tests are working, but the peak memory usage is still around 57GB. I'm not sure what exactly is going on here - running further comprehensive benchmarks should help.


slumber:
Relative commit is [e5cc4b0](https://github.com/nexus-xyz/nexus-zkvm/pull/78/commits/e5cc4b00904931b5fd545f81ba33e9502c6b201d)

Best to pull it into another PR since this one is outdated.

PR #92, opened 2024-02-27
Change `nexus new` from SSH to HTTPS

### LINKED ISSUE #90: `cargo nexus new` fails to run `cargo add --git` command because of SSH auth issue
Doing `cargo nexus new` on a new machine that hasn't setup SSH keys fails because `new` adds the zkvm as a dependency using SSH and not HTTPs.

danielmarinq:
Fixes #90.


dorebell (approved):
lgtm; there's no reason to use SSH for a public repo.

PR #114, opened 2024-02-29
introduce SuperNova

slumber:


govereau (approved):
Very nice. Looks great.

slumber:
@dorebell ping

nova/src/circuits/nova/pcd/mod.rs
```diff
@@ -3,8 +3,8 @@
 //! Similar to [IVC](super::sequential::IVCProof) each [`PCDNode`] proves a range of computing F_i,
 //! with the difference of left index possibly being non-zero.
 //!
-//! A node proving range [i; k) can be folded with another node that proves range [k; j) to build a new
-//! node with range [i; j). These indices correspond to [`PCDNode::min_step`] and [`PCDNode::max_step`].
+//! A node proving range [i; k) can be folded with another node that proves range [k + 1; j) to build
```
dorebell, line 6: good catch!

nova/src/circuits/supernova/augmented.rs
```diff
@@ -0,0 +1,566 @@
...
+            .z_i
+            .iter()
+            .map(|z| FpVar::new_variable(cs.clone(), || Ok(z), mode))
+            .collect::<Result<_, _>>()?;
+
+        let U = input
+            .U
+            .iter()
+            .map(|U| {
+                let U = U.clone().unwrap_or_else(|| RelaxedR1CSInstance {
```
dorebell, line 194: why is this clone necessary? in the nova ivc impl, `U` is moved here.

nova/src/circuits/supernova/augmented.rs
```diff
@@ -0,0 +1,566 @@
...
+        };
+
+        let vk = FpVar::new_variable(cs.clone(), || Ok(input.vk), mode)?;
+        let i = FpVar::new_variable(cs.clone(), || Ok(input.i), mode)?;
+        let pc_prev = FpVar::new_variable(
+            cs.clone(),
+            || Ok(G1::ScalarField::from(input.pc_prev)),
+            mode,
+        )?;
+        let pc = FpVar::new_variable(cs.clone(), || Ok(G1::ScalarField::from(input.pc)), mode)?;
```
dorebell, line 178: we don't need to allocate this variable here: we will compute `pc` in the circuit.

nova/src/circuits/supernova/augmented.rs
```diff
@@ -0,0 +1,566 @@
...
+            },
+        )?;
+
+        for (z_0, z_i) in input.z_0.iter().zip(&input.z_i) {
+            z_0.conditional_enforce_equal(z_i, &is_base_case)?;
+        }
+        input
+            .pc
+            .conditional_enforce_equal(&input.pc_prev, &is_base_case)?;
+        input
```
dorebell, line 403: this constraint is unnecessary. `pc`, which should be computed as _output_ of this circuit, doesn't need to be tied in-circuit to the index of the circuit. this circuit will enforce that `pc_prev` is part of the data hashed into the output of `u`, and fold `u` into `U[pc_prev]`. this then ensures that the index of the circuit for `u` is `pc_prev` (since the final decider will test `U[j]` against the circuit with index `j`)

dorebell (changes requested):

nova/src/circuits/supernova/augmented.rs
```diff
@@ -0,0 +1,566 @@
...
+                    U_secondary,
+                    u,
+                    proof: NIMFSProof::default(),
+                }
+            }
+            SuperNovaAugmentedCircuitInput::NonBase(non_base) => non_base.clone(),
+        };
+
+        let vk = FpVar::new_variable(cs.clone(), || Ok(input.vk), mode)?;
+        let i = FpVar::new_variable(cs.clone(), || Ok(input.i), mode)?;
```
dorebell, line 172:
```suggestion
        let i = FpVar::new_variable(cs.clone(), || Ok(input.i), mode)?;
        let pc_prev = FpVar::new_variable(
```

nova/src/circuits/supernova/augmented.rs
```diff
@@ -0,0 +1,566 @@
...
+    G1::BaseField: PrimeField,
+    G2::BaseField: PrimeField,
+    C1: CommitmentScheme<Projective<G1>>,
+    C2: CommitmentScheme<Projective<G2>>,
+    RO: SpongeWithGadget<G1::ScalarField>,
+{
+    vk: FpVar<G1::ScalarField>,
+    i: FpVar<G1::ScalarField>,
+    pc_prev: FpVar<G1::ScalarField>,
+    pc: FpVar<G1::ScalarField>,
```
dorebell, line 109:
```suggestion
    pc_prev: FpVar<G1::ScalarField>,
```

nova/src/circuits/supernova/mod.rs
```diff
@@ -0,0 +1,631 @@
...
+                (&u, &w),
+            )?;
+
+            let input =
+                SuperNovaAugmentedCircuitInput::NonBase(SuperNovaAugmentedCircuitNonBaseInput {
+                    vk: params.digest,
+                    i: G1::ScalarField::from(i),
+                    pc_prev: pc,
+                    pc: pc_next,
+                    z_0: z_0.clone(),
```
dorebell, line 297:
```suggestion
                    pc_prev: pc,
                    z_0: z_0.clone(),
```

nova/src/circuits/supernova/mod.rs
```diff
@@ -0,0 +1,631 @@
...
+    ) -> Result<Self, cyclefold::Error> {
+        let _span = tracing::debug_span!(
+            target: LOG_TARGET,
+            "prove_step",
+            step_num = %self.step_num(),
+        )
+        .entered();
+        let NIVCProof { z_0, non_base, .. } = self;
+
+        let (i_next, pc, input, U, W, U_secondary, W_secondary) = if let Some(non_base) = non_base {
```
dorebell, line 255:
```suggestion
        let (i_next, _pc, input, U, W, U_secondary, W_secondary) = if let Some(non_base) = non_base {
```

nova/src/circuits/supernova/mod.rs
```diff
@@ -0,0 +1,631 @@
...
+
+            (i_next, pc, input, U, W, U_secondary, W_secondary)
+        };
+
+        let cs = ConstraintSystem::new_ref();
+        cs.set_mode(SynthesisMode::Prove { construct_matrices: false });
+
+        let circuit = SuperNovaAugmentedCircuit::new(&params.ro_config, step_circuit, input);
+
+        let (pc_next, z_i) = tracing::debug_span!(target: LOG_TARGET, "satisfying_assignment")
```
dorebell, line 343:
```suggestion
        let (pc, z_i) = tracing::debug_span!(target: LOG_TARGET, "satisfying_assignment")
```

nova/src/circuits/supernova/mod.rs
```diff
@@ -0,0 +1,631 @@
...
+        let pub_io = cs_borrow.instance_assignment.clone();
+
+        let w = R1CSWitness::<G1> { W: witness };
+
+        let commitment_W = w.commit::<C1>(&params.pp);
+        let u = R1CSInstance::<G1, C1> { commitment_W, X: pub_io };
+
+        let z_i = z_i.iter().map(R1CSVar::value).collect::<Result<_, _>>()?;
+        let pc_next = {
+            let pc_fp = pc_next.value()?;
```
dorebell, line 359:
```suggestion
            let pc_fp = pc.value()?;
```

nova/src/circuits/supernova/mod.rs
```diff
@@ -0,0 +1,631 @@
...
+        let cs = ConstraintSystem::new_ref();
+        cs.set_mode(SynthesisMode::Prove { construct_matrices: false });
+
+        let circuit = SuperNovaAugmentedCircuit::new(&params.ro_config, step_circuit, input);
+
+        let (pc_next, z_i) = tracing::debug_span!(target: LOG_TARGET, "satisfying_assignment")
+            .in_scope(|| {
+                SuperNovaConstraintSynthesizer::generate_constraints(circuit, cs.clone())
+            })?;
+
```
dorebell, line 347:
```suggestion
        
        let (pc_next, z_i) = tracing::debug_span!(target: LOG_TARGET, "satisfying_assignment")
            .in_scope(|| {
                SuperNovaConstraintSynthesizer::generate_constraints(circuit, cs.clone())
            })?;

        let circuit = SuperNovaAugmentedCircuit::new(&params.ro_config, pc_next, step_circuit, input);
```
we can just pass the result of computing the `pc` here, so it doesn't need to be in the `input`.

nova/src/circuits/supernova/augmented.rs
```diff
@@ -0,0 +1,566 @@
...
+    C1: CommitmentScheme<Projective<G1>>,
+    C2: CommitmentScheme<Projective<G2>>,
+    RO: SpongeWithGadget<G1::ScalarField>,
+    SC: NonUniformCircuit<G1::ScalarField>,
+{
+    pub fn new(
+        ro_config: &'a <RO::Var as CryptographicSpongeVar<G1::ScalarField, RO>>::Parameters,
+        step_circuit: &'a SC,
+        input: SuperNovaAugmentedCircuitInput<G1, G2, C1, C2, RO>,
+    ) -> Self {
```
dorebell, line 284:
```suggestion
    pub fn new(
        ro_config: &'a <RO::Var as CryptographicSpongeVar<G1::ScalarField, RO>>::Parameters,
        pc: u64,
        step_circuit: &'a SC,
        input: SuperNovaAugmentedCircuitInput<G1, G2, C1, C2, RO>,
    ) -> Self {
```

nova/src/circuits/supernova/augmented.rs
```diff
@@ -0,0 +1,566 @@
...
+        let pc_value = match &input {
+            SuperNovaAugmentedCircuitInput::Base { pc, .. } => *pc,
+            SuperNovaAugmentedCircuitInput::NonBase(non_base) => non_base.pc,
+        };
+        Self { ro_config, step_circuit, input, pc_value }
```
dorebell, line 289:
```suggestion
        Self { ro_config, step_circuit, input, pc }
```

nova/src/circuits/supernova/mod.rs
```diff
@@ -0,0 +1,631 @@
...
+            return Err(NOT_SATISFIED_ERROR);
+        }
+
+        let mut random_oracle = RO::new(&params.ro_config);
+
+        random_oracle.absorb(&params.digest);
+        random_oracle.absorb(&G1::ScalarField::from(*i));
+        random_oracle.absorb(&G1::ScalarField::from(*pc));
+        random_oracle.absorb(&G1::ScalarField::from(*pc_next));
+        random_oracle.absorb(&self.z_0);
```
dorebell, line 438:
```suggestion
        random_oracle.absorb(&G1::ScalarField::from(*pc));
        random_oracle.absorb(&self.z_0);
```

nova/src/circuits/supernova/mod.rs
```diff
@@ -0,0 +1,631 @@
...
+            U,
+            W,
+            U_secondary,
+            W_secondary,
+            u,
+            w,
+            i,
+            pc,
+            pc_next,
+            z_i,
```
dorebell, line 419:
```suggestion
            pc,
            z_i,
```

dorebell (approved):
looks good overall! having the circuit compute the current program counter instead of the next program counter gets rid of some constraints and matches the conventions in the paper more closely. i don't think it's a crucial change though.

dorebell (approved):
looks good overall! having the circuit compute the current program counter instead of the next program counter gets rid of some constraints and matches the conventions in the paper more closely. i don't think it's a crucial change though.

nova/src/circuits/supernova/augmented.rs
slumber: Because it's either `&input.U[i]` or default instance. Clone can be avoided with `std::borrow::Cow`, but I chose simplicity.

nova/src/circuits/supernova/augmented.rs
slumber: This is a 1 constraint safe guard, your scenario leads to invalid proof as well, but I don't see no harm in it.

nova/src/circuits/supernova/mod.rs
```diff
@@ -0,0 +1,631 @@
...
+}
+
+pub trait NonUniformCircuit<F: PrimeField>: Send + Sync {
+    /// Number of input variables, corresponds to `z.len()`.
+    const ARITY: usize;
+
+    /// Number of computable functions {F_0, ... , F_{l - 1}}.
+    const NUM_CIRCUITS: usize;
+
+    /// Return program counter for the first step.
```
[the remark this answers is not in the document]
slumber: Selector function is part of augmented circuit, which one of `U_i` you're going to execute otherwise?

nova/src/circuits/supernova/mod.rs
```diff
@@ -0,0 +1,631 @@
...
+    /// `generate_constraints()` takes both `pc` and `pc_value` arguments: value of
+    /// the circuit variable `pc` may be missing during setup stage. `pc` is constrained
+    /// to be equal to `pc_value`, so the latter should be used for switching between `F_i`.
+    ///
+    /// Return the next value of the program counter along with output `z_{i + 1}`.
+    fn generate_constraints(
+        &self,
+        cs: ConstraintSystemRef<F>,
+        pc: &FpVar<F>,
+        pc_value: u64,
```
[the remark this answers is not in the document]
slumber: I think it complicates things for trait users a lot if you're not allowed to use `pc` variable.

nova/src/circuits/supernova/mod.rs
```diff
@@ -0,0 +1,631 @@
...
+
+pub trait NonUniformCircuit<F: PrimeField>: Send + Sync {
+    /// Number of input variables, corresponds to `z.len()`.
+    const ARITY: usize;
+
+    /// Number of computable functions {F_0, ... , F_{l - 1}}.
+    const NUM_CIRCUITS: usize;
+
+    /// Return program counter for the first step.
+    fn pc_init(&self) -> u64 {
```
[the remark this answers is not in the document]
slumber: What you suggest is default behavior, I don't see why it's not possible to start with the second augmented circuit or any other.

slumber:
@dorebell the review could've been reduced to several worded-comments on the code, as a lot of code suggestions are related to the same discussion, and there're just too many of them.

nova/src/circuits/supernova/mod.rs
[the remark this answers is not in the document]
slumber: That is, implementors would have to do something like "store the current pc into `z[0]` and switch the pc based on it" -- which will increase `z.len()` and result into the same number of constraints.

nova/src/circuits/supernova/mod.rs
```diff
@@ -0,0 +1,631 @@
...
+            params.shapes[0],
+            params.shape_secondary,
+        );
+        Ok(params)
+    }
+}
+
+pub type PublicParams<G1, G2, C1, C2, RO, SC> =
+    public_params::PublicParams<G1, G2, C1, C2, RO, SC, SetupParams<(G1, G2, C1, C2, RO, SC)>>;
+
```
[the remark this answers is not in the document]
slumber: `clone` impl was useful for benches because of closure captures. Sure no harm adding it

slumber:
> looks good overall! having the circuit compute the current program counter instead of the next program counter gets rid of some constraints and matches the conventions in the paper more closely. i don't think it's a crucial change though.

The paper doesn't pass `pc_prev` into the circuit, so it might only look similar. Lurk's implementation with 2 cycle follows the paper as they only pass current `pc` as the input and fold `u` inside the secondary circuit.

It's not possible with CycleFold because of the base case.

nova/src/circuits/supernova/augmented.rs
```diff
@@ -0,0 +1,566 @@
...
+                let U_secondary = vec![None; *num_augmented_circuits];
+                let u = R1CSInstance::<G1, C1> {
+                    commitment_W: C1::Commitment::default(),
+                    X: vec![G1::ScalarField::ONE; AUGMENTED_CIRCUIT_NUM_IO],
+                };
+                SuperNovaAugmentedCircuitNonBaseInput {
+                    vk: *vk,
+                    i: G1::ScalarField::ZERO,
+                    pc_prev: *pc,
+                    pc: *pc,
```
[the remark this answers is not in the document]
slumber: This change blocks you from generating shapes for different values of pc

dorebell (commented):
> > looks good overall! having the circuit compute the current program counter instead of the next program counter gets rid of some constraints and matches the conventions in the paper more closely. i don't think it's a crucial change though.

In the paper, the circuit input is called `pc_i`. This is the index of the instruction executed by the input `u_i`, and the circuit folds `u_i` into `U_i[pc_i]`. The index-j circuit computes `pc_(i+1)` as a function of `z_i`, and applies `F_j` to `z_i` to get `z_(i+1)`. In IVC, you then set j to be `pc_(i+1)` when deciding which circuit to actually run and record in `u_(i+1)`, but this constraint is not enforced in the circuit. 

The role of `pc_i` in the paper is what you're calling `pc_prev` in the code here, and `pc_(i+1)` is what you're calling `pc`.

dorebell (commented):
> @dorebell the review could've been reduced to several worded-comments on the code, as a lot of code suggestions are related to the same discussion, and there're just too many of them. 

Sorry - there's mainly just one overall suggestion, and i was trying to find all the places the code would change to follow it. would it be easier for me to open a PR against this PR branch with the suggested changes? 

In words, the overall suggestion is that we can remove `pc` from circuit input, as well as all constraints related to it. the circuit does not need to explicitly constrain which step circuit is applied. the circuit should compute `pc` by applying the step circuit, hash it to output, and this will be the `pc_prev` in the next step. 

the circuit constrains that the input `u` is folded into the `pc_prev` instance, which is enough to enforce that the step-circuit computed in `u` is `pc_prev`.

slumber:
> > @dorebell the review could've been reduced to several worded-comments on the code, as a lot of code suggestions are related to the same discussion, and there're just too many of them.
> 
> Sorry - there's mainly just one overall suggestion, and i was trying to find all the places the code would change to follow it. would it be easier for me to open a PR against this PR branch with the suggested changes?

Usually it's PR author who makes changes according to discussions, so I don't see why.

> 
> In words, the overall suggestion is that we can remove `pc` from circuit input, as well as all constraints related to it. the circuit does not need to explicitly constrain which step circuit is applied. the circuit should compute `pc` by applying the step circuit, hash it to output, and this will be the `pc_prev` in the next step.

See https://github.com/nexus-xyz/nexus-zkvm/pull/114#discussion_r1512736630 and https://github.com/nexus-xyz/nexus-zkvm/pull/114#discussion_r1512751722
I don't strongly insist on it though.

As for constraining the circuit to a constant value of pc -- every `U_i` has a different R1CS structure (usually, not necessarily), so it's a natural safeguard, you don't generate constraints for this circuit while the value of `pc` is different.

> the circuit constrains that the input `u` is folded into the `pc_prev` instance, which is enough to enforce that the step-circuit computed in `u` is `pc_prev`.

Makes sense, I'll make according changes.

nova/src/circuits/supernova/mod.rs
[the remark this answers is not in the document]
dorebell: i don't mean first among the set of instructions, i mean the first step in time. the initial `pc` is a function of `z_0`. We require that `F_j(z_0) = (_z, pc_init)` with `pc_init` the same for each choice of `j`. The paper writes this as `pc_init = phi(z_0); z_1 = F_(pc_init)(z_0)`. We could introduce `phi` in the code explicitly, or just note this stipulation.

nova/src/circuits/supernova/mod.rs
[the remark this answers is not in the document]
dorebell: you shouldn't need to know `pc` inside the circuit - just apply the function with index `pc_value` to the input to compute the new state and the return `pc_value`. so like for the precompiles, we'd presumably always return the VM's `pc_value` as the next `pc`

i think having the entire state of the VM encapsulated by `z` is a good invariant to maintain.

nova/src/circuits/supernova/mod.rs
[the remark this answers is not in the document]
slumber: 1. This would require implementors to be additionally careful with `NonUniformCircuit` trait.

2. `phi` can only be executed as part of some augmented circuit (say we don't care which one because of \all j requirement)
Does it mean that in the base case you first compute instanse `u'` that outputs `pc_init`, and then pass it as an input to the actual base step to get `u`? Or do you discard `u'`?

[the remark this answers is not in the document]
dorebell: I suggest something like 
```
pub trait NonUniformCircuit<F: PrimeField>: Send + Sync {
    /// Number of input variables, corresponds to `z.len()`.
    const ARITY: usize;

    /// Number of computable functions {F_0, ... , F_{l - 1}}.
    const NUM_CIRCUITS: u64;

    /// Generate constraints for computing the selector function "phi".
    ///
    /// Return the next value of the program counter.
    fn generate_constraints_pc(
        &self,
        cs: ConstraintSystemRef<F>,
        i: &FpVar<F>,
        z: &[FpVar<F>],
    ) -> Result<FpVar<F>, SynthesisError>;

    /// Generate constraints for applying function `F_(pc_value)` to `z`.
    ///
    /// Return the next value of `z`, `z_(i+1)`.
    fn generate_constraints_step(
        &self,
        cs: ConstraintSystemRef<F>,
        pc_value: u64,
        i: &FpVar<F>,
        z: &[FpVar<F>],
    ) -> Result<Vec<FpVar<F>>, SynthesisError>;

    /// Generate constraints for applying both `phi` and `F_{pc_value}` to `z`.
    ///
    /// Returns the next value of the program counter and the next value of `z`, `z_(i+1)`.
    /// The next value of the program counter is guaranteed not to depend on `pc_value`.
    fn generate_constraints(
        &self,
        cs: ConstraintSystemRef<F>,
        pc_value: u64,
        i: &FpVar<F>,
        z: &[FpVar<F>],
    ) -> Result<(FpVar<F>, Vec<FpVar<F>>), SynthesisError> {
        let pc = self.generate_constraints_pc(cs, i, z)?;
        let z_next = self.generate_constraints_step(cs, pc_value, i, z)?;
        Ok((pc, z_next))
    }
}
```

then in the `SupernovaAugmentedCircuit::generate_constraints()` you could have something like

``` 
    let pc = <SC as NonUniformCircuit>::generate_constraints_pc(circuit, cs, i, z)?;
    let pc_value: u64 = pc.value().into().unwrap();
    let z_next = <SC as NonUniformCircuit>::generate_constraints_step(circuit, cs, pc_value, i, z)?;
```
the relationship between `pc` and `pc_value` is _not_ enforced by the circuit, but just by the code. 

I think this is cleanest, but to be closer to the paper, you could have `generate_constraints()` take `pc_value` as input. then in a proof step, the prover first computes `pc_value = <SC as NonUniformCircuit>::generate_constraints_pc(circuit, dummy_cs, i, z)`, throwing out the constraint system/instance/witness, then uses this in the call to `generate_constraints()`.

nova/src/circuits/supernova/mod.rs
```diff
@@ -501,33 +528,44 @@ pub(crate) mod tests {
...
+            assert_eq!(z.len(), 2);
 
-            let pc_is_zero = pc.is_zero()?;
-            let pc_next = pc_is_zero.select(&FpVar::one(), &FpVar::zero())?;
-
-            let x = &z[0];
+            let x = &z[1];
 
-            let y = match pc_value {
+            let y = match pc {
```
dorebell, line 557: this looks a bit like collatz... :)

nova/src/circuits/supernova/mod.rs
```diff
@@ -501,33 +528,44 @@ pub(crate) mod tests {
     pub struct TestCircuit<F: Field>(PhantomData<F>);
 
     impl<F: PrimeField> NonUniformCircuit<F> for TestCircuit<F> {
```
dorebell, line 530: this looks good! i think having the circuit compute `pc_next` without being given `pc` directly will be more natural in the actual VM context than it is here. since there `z` already is a commitment to a whole bunch of values (memory etc), including the risc-v pc. i'm imagining that we'd have something like `pc = 0` for the VM circuit, and precompiles would probably just always return `0` for `compute_selector()` and the VM circuit would derive the new `pc` from the program stored in memory.

dorebell (approved):
Perfect! Thanks for digging through the trenches with me on this one. it's probably not a crucial change, but staying as close to the paper as possible helps convince people that the soundness results in the paper apply. (i'm 99% sure your implementation constrained everything enough to retain soundness, but to be totally rigorous with it we'd want to write out the security proof). 

all remaining comments are just typos or similarly minor nits, nothing important.

nova/src/circuits/supernova/augmented.rs
```diff
@@ -397,26 +402,36 @@ where
...
+        let pc_next = <SC as NonUniformCircuit<G1::ScalarField>>::compute_selector(
             self.step_circuit,
             cs.clone(),
-            &input.pc,
-            self.pc_value,
+            &input.i,
+            &input.z_i,
+        )?;
+        let pc_value = self.pc_hint.unwrap_or_else(|| {
+            let pc_next = pc_next.value().expect("hint must be provided for cs setup");
```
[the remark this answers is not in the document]
slumber: If this code panics it means incorrect usage of the internal interface
https://github.com/nexus-xyz/nexus-zkvm/blob/d61faeeecc41ecd02a92b556acb550b732122087/nova/src/circuits/supernova/augmented.rs#L373

nova/src/circuits/supernova/mod.rs
slumber: it was renamed in the trait definition :man_shrugging:

nova/src/circuits/supernova/augmented.rs
```diff
@@ -35,6 +35,22 @@ pub const SQUEEZE_NATIVE_ELEMENTS_NUM: usize = 1;
 /// Leading `Variable::One` + 1 hash.
 pub const AUGMENTED_CIRCUIT_NUM_IO: usize = 2;
 
+/// Converts prime field element into u64.
+///
+/// Panics if it doesn't fit into 8 bytes.
+pub(super) fn field_to_u64<F: PrimeField>(f: F) -> u64 {
```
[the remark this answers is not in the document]
slumber: If the circuit returns value out of bounds (and especially greater than u64::MAX) it's better to just panic: step circuit implementation has fundamental problems, i.e. you can't really handle such error. Just like indexing into array and going out of bounds.

PR #126, opened 2024-03-07
abstract over memory implementation

### LINKED ISSUE #125: introduce memory controller
Introduce the notion of a "memory controller" that is responsible for providing a backing store, reading and writing cache-lines, and generating proof objects.

govereau:
This patch introduces the concept of a "memory controller" which is responsible for providing a backing store, and generating commitments and proof objects.

closes #125


slumber:
How hard would it be to make memory implementation a part of `VmConfig` and allow switching it with env variables?

It's not something we need to implement right now, but it could be a nice-to-have for benches and tests.

govereau:
> How hard would it be to make memory implementation a part of `VmConfig` and allow switching it with env variables?
> 
> It's not something we need to implement right now, but it could be a nice-to-have for benches and tests.

Should be pretty easy, I think. We definitely want to do this. Probably in a future PR where we can also handle the IO and syscall configurations.

slumber (approved):

PR #128, opened 2024-03-19
upload benches results and update readmes

slumber:
Additionally update the bench itself and check that it compiles in CI routines.


danielmarinq:
Looks great!

danielmarinq (approved):

PR #130, opened 2024-03-22
Updated Nexus proof system section in docs

michel-nexus:


docs/pages/Specs/nexus-prover.mdx
```diff
@@ -35,32 +35,124 @@ After receiving a Merkle proof from the initialization step with a Merkle proof
...
+
+Besides not having to reason about the 2-cycle of elliptic curves in the security proof of Nova [[KS23](#references)], one important advantage of CycleFold's approach over the one used in Microsoft's reference implementation is that the circuit size over the secondary curve becomes significantly smaller. More concretely, the size of the IVC verifier of the Nexus zkVM on the secondary curve is in the order of $3{,}000$ constraints.
+
+The reduction in circuit complexity over the secondary curve has, however, implications on the size of the IVC verifier over the primary curve since the latter circuit now needs to keep track of additional witnesses and to perform several additional small multi-scalar multiplications and hashing operations related to these witnesses. More concretely, the recursion overhead over the primary curve is in the order of $70{,}000$ constraints when used to realize an IVC scheme and $340{,}000$ constraints when used to realize a PCD scheme.
```
slumber, line 66: Sequential Nova requires 120k constraints, 70k is achievable with emulated arithmetic optimization.

docs/pages/Specs/nexus-prover.mdx
```diff
@@ -35,32 +35,124 @@ After receiving a Merkle proof from the initialization step with a Merkle proof
...
+
+Hence, unlike Microsoft's reference implementation, the Nexus zkVM follows the approach proposed by CycleFold [[KS23](#references)] in its implementation of Nova to address concerns about non-native elliptic-curve arithmetic operations in later phases of the proof system. In this approach, instead of duplicating the IVC chain in both curves, the circuit implemented over the second curve is only used to perform a single group scalar multiplication over the base field of the first curve. Moreover, this is done in a way that allows the verifier in the first curve to avoid non-native arithmetic operations by verifiably delegating its group scalar multiplications to the second curve.
+
+Besides not having to reason about the 2-cycle of elliptic curves in the security proof of Nova [[KS23](#references)], one important advantage of CycleFold's approach over the one used in Microsoft's reference implementation is that the circuit size over the secondary curve becomes significantly smaller. More concretely, the size of the IVC verifier of the Nexus zkVM on the secondary curve is in the order of $3{,}000$ constraints.
```
slumber, line 64: We only need around 1500 for sequential case, because you only need 128 bits computed for scalar multiplication.

But secondary circuit implementation is unified for PCD and IVC. And PCD does need 3k since `r^2` isn't guaranteed to be a 128 bit number.

Your docs are correct, but you can add this as `>` note.

docs/pages/Specs/nexus-prover.mdx
```diff
@@ -35,32 +35,124 @@ After receiving a Merkle proof from the initialization step with a Merkle proof
...
+
+In a nutshell, Microsoft's reference implementation for Nova works by creating two parallel copies of the IVC chain linked together in a way that allows the recursive step of one IVC chain to be verified on the other chain using native elliptic curve arithmetic operations. As shown in [[KST22](#references)], this results in a largely symmetric scheme with a recursion overhead in the order of $20{,}000$ constraints in each of the curves. Though the Nova reference implementation can work over most cycles of curves, its current implementation uses the [`pasta` cycle](https://github.com/zcash/pasta_curves) of two curves.
```
slumber, line 52: 1. Microsoft Nova overhead is around 10000
2. They support both pasta and grumpkin cycles

docs/pages/Specs/nexus-prover.mdx
```diff
@@ -35,32 +35,124 @@ After receiving a Merkle proof from the initialization step with a Merkle proof
...
+const PARTIAL_ROUNDS: usize = 57;
+const ALPHA: u64 = 5;
+const RATE: usize = 2;
+const CAPACITY: usize = 1;
+const MODULUS_BITS: u64 = 254;
+```
+
+### Instructions per folding step
+
+The Nexus zkVM's approach based on Nova provides a proof system with a recursion overhead which is independent of the total number of computation steps and of the step function itself. However, in our current implementation, the recursion overhead  remains quite high (in the order of $70{,}000$ and $340{,}000$ constraints in the IVC and PCD settings) compared the cost of a single step of the Nexus VM (around $15{,}000$ constraints).
```
slumber, line 110: same comment on the IVC overhead

govereau (approved):

docs/pages/Specs/nexus-prover.mdx
michel-nexus: Regarding (1), I was using the number from Figure 1 in [KST22], but these numbers seem to be outdated so I will update the text. Would you have a reference for the latest numbers for the recursion overhead for Microsoft Nova?

Regarding (2), I'm adding this remark to the text.

docs/pages/Specs/nexus-prover.mdx
michel-nexus: Thank you for the clarification. I'm adding a note to the text.

docs/pages/Specs/nexus-prover.mdx
michel-nexus: I'm updating the value of this he IVC overhead in this he text and adjusting the value of k in the example.

docs/pages/Specs/nexus-prover.mdx
michel-nexus: I'm fixing the IVC overhead value in the text.

docs/pages/Specs/nexus-prover.mdx
michel-nexus: Added a reference for the Microsoft Nova IVC overhead

PR #138, opened 2024-04-02
Change Default Build Options

sjudson:
This PR implements a new build profile, `release-unoptimized`, which is inserted in the `Cargo.toml` of any new nexus project as well as used for the `examples` folder of the `nexus-zkvm` repo. This new profile is equivalent to `release` except with `opt-level = 0`.

It then:

a) sets the default profile for `cargo nexus run` to `dev`,
b) sets the default profile for `cargo nexus prove` to `release-unoptimized`,
c) enables a `--profile` option for both commands.


tools/tools-dev/src/command/common/run.rs
```diff
@@ -6,10 +6,14 @@ pub struct RunArgs {
     #[arg(short)]
     pub verbose: bool,
 
-    /// Build artifacts in release mode, with optimizations.
-    #[arg(short, long)]
+    /// Build artifacts with the release profile. Equivalent to "--profile release".
+    #[arg(short, name = "r", conflicts_with = "profile")]
```
slumber, line 10: Why do you think `--profile=release` isn't sufficient? This is also reflected in the code as it conflicts.

`--profile` is explicit, `--release` for usual cargo is for distinction between debug and optimized builds, in our case it's more involved.

slumber: Same with prove command

sjudson: Was just looking to retain parity with the previous args (at least for `cargo nexus run`). Happy to remove if we want the user to be fully explicit.

govereau: I guess I agree it should be disabled by default if we think there is a good chance the translation will not work with release. Probably we should go back to a 32-bit encoding so the we don't have this issue.

tools/tools-dev/src/command/common/new.rs
```diff
@@ -32,6 +32,20 @@ fn setup_crate(path: PathBuf) -> anyhow::Result<()> {
...
     )?;
 
+    let mut fp = fs::OpenOptions::new()
+        .append(true)
+        .open(path.join("Cargo.toml"))?;
+
+    writeln!(
+        fp,
+        concat!(
+            "\n",
```
slumber, line 42: Let's add a comment here, something like
```toml
# generated by cargo-nexus, do not remove!
#
# this profile is used for generating proofs and running the program, 
# currently nexus-vm doesn't handle compiler optimizations correctly. 
[profile.release-unoptimized]
```

slumber: (feel free to word it yourself, I'm not good at this)

Issue #76, opened 2024-02-17
Implement HyperNova


sjudson:
- [x] Implement CCS (https://github.com/nexus-xyz/nexus-zkvm/pull/52)
- [x] Implement NIMFS (https://github.com/nexus-xyz/nexus-zkvm/pull/124)
- - [x] Modify multilinear sumcheck implementation imported from Arkworks (https://github.com/nexus-xyz/nexus-zkvm/commit/c8fe9df425211134db250ef40f395cd12fb49322) as needed  
- [x] Implement IVC (https://github.com/nexus-xyz/nexus-zkvm/pull/144)
- - [x] Implement folding verification gadget for R1CS-derived HyperNova, including in-circuit sumcheck (https://github.com/nexus-xyz/nexus-zkvm/pull/141)

PR #141, opened 2024-04-05
HyperNova Gadgets

sjudson:
Working towards: https://github.com/nexus-xyz/nexus-zkvm/issues/76

This PR implements HyperNova (CycleFold) folding gadgets for constraints derived from R1CS.


nova/src/gadgets/cyclefold/hypernova/primary.rs
```diff
@@ -0,0 +1,561 @@
...
+        Self {
+            commitment_W: self.commitment_W.clone(),
+            X: self.X.clone(),
+            _commitment_scheme: self._commitment_scheme,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct CCSInstanceFromR1CSVar<G1, C1>(CCSInstanceVar<G1, C1>)
```
sjudson, line 52: I wanted to make the implementation explicit about the necessary source of the constraints, as well as to interface with `ark-r1cs-std::R1CSVar` in a clear way, and this seemed like the most minimal way to do so. Happy to consider other approaches though.

nova/src/ccs/mod.rs
```diff
@@ -302,6 +331,17 @@ impl<G: CurveGroup, C: PolyCommitmentScheme<G>> LCCSInstance<G, C> {
         }
     }
 
+    /// A method to create a default instance object with an appropriate shape for the base case of IVC.
+    pub fn base(shape: &CCSShape<G>) -> Self {
```
slumber, line 335: I don't think you will have `shape` available in the base case of the ivc, unless you either introduce lifetimes everywhere and borrow it.

Seems irrelevant for this PR, but also insignificant.

slumber (approved):
Looks great!

Absorbing LCCS instance with a huge step circuit might increase the recursion overhead (`rs.len()` grows) by several thousands, but the multifold itself has 30% less constraints than original Nova.

nova/src/ccs/mod.rs
sjudson: Ah this was a part of sketching out the IVC and I forgot to cut it from this PR along with the rest. I'll remove it.

PR #143, opened 2024-04-11
revert to using rng in Pedersen::setup

slumber:
also fixes tools and reverts accident commented example merge

Public parameters tested to take almost the same amount of time as with hashtocurve.
Reverted commit is available in git/pr history if we ever need it.


slumber:
(review 2 commits after the revert)

PR #144, opened 2024-04-17
HyperNova IVC

sjudson:
Completing https://github.com/nexus-xyz/nexus-zkvm/issues/76, this PR implements HyperNova (CycleFold) IVC, including benchmarking.


nova-benches/benches/recursive-snark.rs
```diff
@@ -39,63 +48,130 @@ criterion_group! {
...
+                .verify(black_box(&pp), black_box(NUM_WARMUP_STEPS))
+                .unwrap();
         });
+    });
+
+    group.finish();
+}
+
+#[cfg(feature = "spartan")]
+fn hypernova(c: &mut Criterion, num_cons_in_step_circuit: usize) {
```
sjudson, line 109: Reminder to myself to factor out into a separate file.

nova/src/ccs/mod.rs
```diff
@@ -72,6 +72,21 @@ pub struct CCSShape<G: CurveGroup> {
     pub cSs: Vec<(G::ScalarField, Vec<usize>)>,
 }
 
+impl<G: CurveGroup> fmt::Display for CCSShape<G> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "CCSShape {{ num_constraints: {}, num_vars: {}, num_io: {}, num_matrices: {}, num_multisets: {}, max_cardinality: {}, Ms: [_, {}], cSs: [_, {}] }}",
```
slumber, line 77: nit `A: [_, {N}]` in R1CS means that A is a matrix with `N` non-zero elements. Here it's `Ms: [_, {number of M matrices}]`

sjudson: I mean this is mostly a gut check anyway, the most relevant size information is captured in the first three parameters

nova/src/lib.rs
```diff
@@ -21,9 +21,10 @@ pub mod commitment;
 pub mod r1cs;
 
 pub use circuits::{
+    hypernova::{self}, // uses same StepCircuit trait as Nova
```
slumber, line 24: How's clippy happy though

nova/src/circuits/hypernova/sequential/mod.rs
```diff
@@ -0,0 +1,590 @@
...
+        let mut rng = rand_chacha::ChaCha20Rng::from_seed(seed);
+
+        let max_poly_vars = ((shape.num_vars.max(shape.num_constraints) - 1)
+            .checked_ilog2()
+            .unwrap_or(0)
+            + 1) as usize;
+
+        let srs = C1::setup(max_poly_vars, b"hypernova_seq_primary_curve", &mut rng)
+            .map_err(|_| cyclefold::Error::PolyCommitmentSetup)?;
+
```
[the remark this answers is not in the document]
sjudson: Ah I see, take in the SRS as SetupAux, will do.

nova/src/provider/zeromorph.rs
```diff
@@ -0,0 +1,108 @@
+use std::marker::PhantomData;
```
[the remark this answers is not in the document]
sjudson: I could have, but it has its own logging that I didn't want to stomp on. Plus integrating the SRS from SetupAux stuff will add concrete functionality that makes sense here instead of Spartan (as in the pcd/compression code).

nova/src/circuits/hypernova/sequential/mod.rs
```diff
@@ -0,0 +1,590 @@
...
+    pub fn verify(
+        &self,
+        params: &PublicParams<G1, G2, C1, C2, RO, SC>,
+        num_steps: usize,
+    ) -> Result<(), cyclefold::Error> {
+        let _span = tracing::debug_span!(target: LOG_TARGET, "verify", %num_steps).entered();
+
+        const NOT_SATISFIED_ERROR: cyclefold::Error =
+            cyclefold::Error::Ccs(crate::ccs::Error::NotSatisfied);
```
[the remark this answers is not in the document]
sjudson: IIRC clippy didn't like it for some reason, but I can take another look + potentially just tell clippy to allow.

nova/src/provider/zeromorph.rs
```diff
@@ -91,7 +94,21 @@ where
         )
         .entered();
 
-        SpartanZM::setup(max_poly_vars, label, rng)
```
slumber: Why is this change needed?

In practice it's not going to be used as SRS requires trusted setup.

sjudson: Just trying to make test setup as low touch as possible

sjudson: I guess I can push this elsewhere though, will change shortly

slumber: Okay I mean why does this code **need** to be somewhere.
This kind of matters for Pedersen bases, for tests you can use test_rng and it's good enough without additional hashing and seeds.

sjudson: I guess that's true, okay I can strip it down to the `test_rng`.

nova/src/utils.rs
```diff
@@ -35,6 +35,13 @@ pub fn iter_bits_le(bytes: &[u8]) -> impl Iterator<Item = bool> + '_ {
         .flat_map(|byte| (0..8).map(move |bit| ((1 << bit) & byte) != 0))
 }
 
+#[macro_export]
+macro_rules! safe_log {
+    ($x:expr) => {
+        ($x - 1).checked_ilog2().unwrap_or(0) + 1
+    };
+}
+
```
slumber, line 44: 1. It's good to add a doc-comment explaining what it does: computes the number of bits needed to represent numbers in range
[0; x)
For example, given an array of N elements you need this many bits to represent each index.
2. It's not really safe:
```rust
let i = std::hint::black_box(0u64);
println!("{}", safe_log!(i));
```
prints 64 in release. I don't think we ever use it with zeros (it's asserted in most places), but it's either the name or the underflow check.

slumber (approved):

nova/src/utils.rs
sjudson: Done.

nova/src/utils.rs
```diff
@@ -35,10 +35,17 @@ pub fn iter_bits_le(bytes: &[u8]) -> impl Iterator<Item = bool> + '_ {
         .flat_map(|byte| (0..8).map(move |bit| ((1 << bit) & byte) != 0))
 }
 
+/// An implementation of base_2 log that will never fail, instead returning 0 for all values smaller than 1.
+/// It is 'safe' in the sense that it will not panic, unlike just calling `.unwrap()` on a `.checked_ilog2`.
+/// Usually used to compute the number of bits needed to index a set or a vector.
```
slumber, line 40: It should mention that it returns ceil of log2 then, "safety" isn't the only difference from std ilog2, with a special case of `safe_log!(1) = 1`

nova/src/utils.rs
```diff
@@ -35,10 +35,17 @@ pub fn iter_bits_le(bytes: &[u8]) -> impl Iterator<Item = bool> + '_ {
...
 #[macro_export]
 macro_rules! safe_log {
     ($x:expr) => {
-        ($x - 1).checked_ilog2().unwrap_or(0) + 1
+        match $x {
+            x if x <= 1 => 0,
```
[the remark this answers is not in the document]
slumber: Although seeing that the code of mle.rs changed I'm not sure that we care about x = 1

nova/src/utils.rs
```diff
@@ -35,15 +35,21 @@ pub fn iter_bits_le(bytes: &[u8]) -> impl Iterator<Item = bool> + '_ {
...
+///   log_2(x) = y for x > 1
+///              1 for x = 1
+///              0 otherwise
+///
 /// It is 'safe' in the sense that it will not panic, unlike just calling `.unwrap()` on a `.checked_ilog2`.
-/// Usually used to compute the number of bits needed to index a set or a vector.
+/// Used to compute the number of bits needed to index a set or a vector.
 #[macro_export]
-macro_rules! safe_log {
+macro_rules! safe_loglike {
```
slumber, line 48: `ceil_log2` maybe?

sjudson: It's not quite `ceil` though, and neither is it `clamped` because `log_2 = 0` and we're redefining that.

PR #145, opened 2024-04-18
Dynamic Memory Limit

sjudson:
This PR allows dynamically setting the memory of the runtime through an (optional) argument to the `#[nexus_rt::main]` macro, _e.g._, `#[nexus_rt::main(memlimit(16))]` sets a 16mb limit.


examples/src/bin/memset.rs
```diff
@@ -0,0 +1,19 @@
...
+
+// use 16mb
+// alternatively, remove memset argument or otherwise set less than (slightly more than) 6mb and it will fail
+#[nexus_rt::main(memset(16))]
+fn main() {
+    let vec = vec![0 as u32; 1500000];
+    println!(
+        "Arbitrary vector entry so that it does not get optimized away: vec[42] = {}",
+        vec[42]
+    );
```
slumber, line 17: https://doc.rust-lang.org/core/hint/fn.black_box.html ?

slumber (commented):
What does it actually configure? At a glance it's the stack size, but "memset" usually means setting some memory region to a given value.

sjudson:
> What does it actually configure? At a glance it's the stack size, but "memset" usually means setting some memory region to a given value.

Ah fair point, yeah let me use a different name.

govereau (approved):

PR #26, opened 2023-12-20
sketch of nexus API

govereau:

PR #146, opened 2024-04-19
Nexus API

sjudson:
This PR supplants https://github.com/nexus-xyz/nexus-zkvm/pull/26, with a more opinionated approach about API structure and a few more examples.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ members = [
     "network",
     "nova",
     "spartan",
+    "api",
 ]
 default-members = [
     "riscv",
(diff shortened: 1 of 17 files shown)


slumber (approved):
The crate itself re-exports functionality already available in `nexus-*` crates, the only new thing it has compared to https://github.com/nexus-xyz/nexus-zkvm/blob/slumber-script-example/prover/examples/prove.rs is an input tape.

Won't argue since it looks like it was planned and agreed on, but there's IMO.

govereau (approved):
looks good

PR #149, opened 2024-04-22
Fix Register Passing for Syscalls

sjudson:
Pass registers by mutable reference for syscalls to allow private input tape reading/writing.


sjudson:
Supplanted by https://github.com/nexus-xyz/nexus-zkvm/pull/150

PR #151, opened 2024-04-30
Bump version to 1.0.0

sjudson:
Also has quickstart docs reference the stable version.


mx00s:
LGTM, but perhaps an additional approver would be a good idea for external version bump PRs.

mx00s (approved):

slumber (approved):

PR #159, opened 2024-05-20
integrate Jolt as default prover

slumber:


slumber:
How to test the PR:
```sh
git fetch && git switch slumber-integrate-jolt
```
Then copy the path to the workspace, since tools use the gh version of runtime we'll need to replace it:
```sh
pwd # save the output
cargo install --path tools
```

Create a new nexus project outside of workspace (as usual) and replace the line in Cargo.toml
```toml
# nexus-rt = { git = "https://github.com/nexus-xyz/nexus-zkvm.git", version = "1.0.0" }
nexus-rt = { path = "{path to nexus}/runtime" }
``` 

### Prove
Tools use jolt by default:
```sh
cargo nexus prove
cargo nexus verify
```

To use Nova do either
```sh
cargo nexus prove --impl=nova-seq
cargo nexus verify --impl=nova-seq
```
or
```sh
export NEXUS_VM_PROVER=nova-seq
cargo nexus prove
cargo nexus verify
```

> NOTE: prints are not currently supported, as well assertions, both involve using unsupported ecalls.

slumber:
I've also left a [README](https://github.com/nexus-xyz/nexus-zkvm/tree/slumber-integrate-jolt/runtime/linker-scripts) describing the linker scripts hack. Once we remove nvm translation we should switch to using a single script. Jolt can be modified for a smaller memory shift, ~2**7 should be enough and insignificant.

danielmarinq:
This looks great. I'm getting some nasty error about `cargo` for the `1.77.0-aarch64-apple-darwin` toolchain. Which toolchain are you using?

Screenshot 2024-05-20 at 9 52 04 PM

slumber:
> This looks great. I'm getting some nasty error about `cargo` for the `1.77.0-aarch64-apple-darwin` toolchain. Which toolchain are you using?
> Screenshot 2024-05-20 at 9 52 04 PM

I'll test it on mac, works for me on linux

slumber:
@danielmarinq seems like unrelated issue, can you try https://stackoverflow.com/questions/76307792/cannot-find-cargo-binary-on-rust-toolchain-on-apple-m1?

mx00s:
My blackbox test on an M3 worked. @danielmarinq, in case it helps, here's what `rustup show` says for me in the path to the generated Nexus crate:

```
$ rustup show
Default host: aarch64-apple-darwin
rustup home:  <$HOME>/.rustup

installed toolchains
--------------------

stable-aarch64-apple-darwin (default)
1.77.0-aarch64-apple-darwin

installed targets for active toolchain
--------------------------------------

aarch64-apple-darwin
riscv32i-unknown-none-elf

active toolchain
----------------

1.77.0-aarch64-apple-darwin (overridden by '/private/tmp/test-pr159/rust-toolchain.toml')
rustc 1.77.0 (aedd173a2 2024-03-17)
```

# Basic performance comparison

This is based on the default crate generated from `cargo nexus new <PATH>` with no changes other than using my local clone (rev: 02b962e) of the runtime as @slumber described. I used this small script to keep everything consistent:

```shell
#!/bin/bash

set -e

ZK_IMPL=$1

echo "cleaning ..."
cargo clean

echo "pre-compiling prover '${ZK_IMPL}'"
cargo nexus prove --impl "${ZK_IMPL}" 2>/dev/null 1>/dev/null
rm -f nexus-proof

echo "proving and verifying using '${ZK_IMPL}'"
time cargo nexus prove --impl "${ZK_IMPL}"
time cargo nexus verify --impl "${ZK_IMPL}"
```

## jolt

Output from `./run.sh jolt`:

```
cleaning ...
     Removed 70 files, 5.0MiB total
pre-compiling prover 'jolt'
proving and verifying using 'jolt'
    Finished release-unoptimized [unoptimized] target(s) in 0.00s
  Preprocessing bytecode ... 481ms
       Finished in 481ms; bytecode size: 150
Executing program...
Executed 1338 instructions in 324.208µs
  Proving program execution ... 2.2s
 Finished in 2.2s
  Saving proof ... 3ms
Finished in 3ms

real	0m2.697s
user	0m12.846s
sys	0m1.888s
  Preprocessing bytecode ... 468ms
       Finished in 468ms; bytecode size: 150
  Verifying proof ... 164ms
   Finished in 164ms

real	0m0.680s
user	0m1.136s
sys	0m0.187s
```

## nova-seq

Output from `./run.sh nova-seq`:

```
cleaning ...
     Removed 96 files, 19.6MiB total
pre-compiling prover 'nova-seq'
proving and verifying using 'nova-seq'
    Finished release-unoptimized [unoptimized] target(s) in 0.00s
  Loading public parameters ... 3.3s
 Finished in 3.3s
  Computing step 0 ... 507ms
<snip>
  Computing step 126 ... 717ms
     Proved 127 step(s) in 1min31s; 22.22 instructions / second
  Saving proof ... 17ms
Finished in 17ms

real	1m34.953s
user	8m7.791s
sys	0m16.333s
  Loading public parameters ... 3.3s
 Finished in 3.3s
  Verifying proof ... 622ms
   Finished in 622ms

real	0m3.963s
user	0m8.598s
sys	0m0.153s
```

danielmarinq:
Great, it seems we now need unified logging. @sjudson should be able to handle this (if deemed important).

mx00s:
> NOTE: prints are not currently supported, as well assertions, both involve using unsupported ecalls.

Got it.

Here's how that manifests in the UX:

# Assertions

With this minor change to the default generated crate ...

```diff
diff --git a/src/main.rs b/src/main.rs
index 2162d2f..4856c23 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -14,5 +14,6 @@ fn main() {
         }
     }
     let n = core::hint::black_box(12);
-    let _result = f(n);
+    let result = f(n);
+    assert_eq!(result, 479001600);
 }
```

```shell
$ cargo nexus prove --impl jolt
   Compiling test-pr159 v0.1.0 (/private/tmp/test-pr159)
    Finished release-unoptimized [unoptimized] target(s) in 0.03s
Error: Instruction isn't supported: ecall
```

# Prints

```diff
diff --git a/src/main.rs b/src/main.rs
index 2162d2f..c9ce194 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,16 +3,5 @@
 
 #[nexus_rt::main]
 fn main() {
-    fn f(n: u32) -> u32 {
-        if n <= 1 {
-            1
-        } else {
-            // n * f(n - 1) would panic if the factorial overflows u32::MAX in debug build,
-            // and wrap around in release. Therefore, use built-in checked methods to make
-            // the output deterministic.
-            n.saturating_mul(f(n - 1))
-        }
-    }
-    let n = core::hint::black_box(12);
-    let _result = f(n);
+    nexus_rt::write_log("Hello, world!\n");
 }
```

```shell
$ cargo nexus prove --impl jolt
   Compiling test-pr159 v0.1.0 (/private/tmp/test-pr159)
    Finished release-unoptimized [unoptimized] target(s) in 0.03s
Error: Instruction isn't supported: ecall
```

jolt/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+    pub fn bytecode_size(&self) -> usize {
+        self.insts.len()
+    }
+}
+
+pub fn preprocess(vm: &VM) -> JoltPreprocessing {
+    rv32i_vm::RV32IJoltVM::preprocess(
+        vm.insts.clone(),
+        vm.mem_init.clone(),
+        1 << 22,
```
sjudson, line 63: It would be great if we could have some descriptive comments here for any of the magic numbers that pop up as part of converting the Jolt API. Will make it easier to maintain the interface in the future.

jolt/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+
+pub fn verify(
+    preprocessing: JoltPreprocessing,
+    proof: JoltProof,
+    commitments: JoltCommitments,
+) -> Result<(), Error> {
+    rv32i_vm::RV32IJoltVM::verify(preprocessing, proof, commitments).map_err(Into::into)
+}
+
+fn build_jolt_trace<F: JoltField>(
```
sjudson, line 94: This function is surprisingly clean and straightforward to me, very positive for us being able to easily maintain this interface going forward.

jolt/src/trace.rs
```diff
@@ -0,0 +1,129 @@
...
+/// Trace VM execution in Jolt format.
+pub fn trace(vm: VM) -> Result<Vec<jolt_rv::RVTraceRow>, Error> {
+    let VM { mut vm, insts, .. } = vm;
+
+    let mut trace = Vec::new();
+    loop {
+        // decode next inst
+        let slice = vm.mem.load(Width::W, vm.regs.pc)?.0.to_le_bytes();
+        let next_inst = parse_inst(vm.regs.pc, &slice)?;
+        if next_inst.inst == RV32::UNIMP {
```
sjudson, line 20: Shouldn't we scan ahead to see if there are any non-unimp instructions following this point, and error if so? 

Feels like we should fail closed instead of fail open here, so that we're not proving partial programs, even in degenerate cases.

jolt/src/trace.rs
```diff
@@ -0,0 +1,129 @@
...
+                };
+
+                let x = vm.get_reg(rs1);
+                let addr = add32(x, imm);
+                Some((width, addr))
+            } else {
+                None
+            };
+
+        let inst = convert::inst(next_inst);
```
sjudson, line 41: Another code block here that I think is pleasingly simple and straightforward.

runtime/build.rs
```diff
@@ -5,6 +5,18 @@ use std::fs;
 use std::path::PathBuf;
 
 fn main() {
+    // START dynamic linker script hack
```
sjudson, line 8: Yeah this isn't great but could definitely be a lot worse.

sjudson (changes requested):
Overall I think this looks really good. I just have one substantive comment (about defensively handling the presence of an `UNIMP` instruction in the trace).

The rest is just about documenting some of the conversions a bit better and my pleasant surprise about how non-invasive this is.

As an overarching comment though, I think it would probably be best to reconfigure this PR to not change the default prover. I can then come through and add ECALL support (it shouldn't be that hard) as well as change the default.

jolt/src/trace.rs
slumber: It's unlikely you'll succeed decoding anything beyond this point, and this is exactly how current vm works -- stops executing if the decoded instruction is `UNIMP`.

sjudson: Sort of? At the moment we canonically translate `RV32::UNIMP` to `HALT` in the NVM, so unambiguously the semantics of the compiled program would not execute past the first appearance of `RV32::UNIMP`/`HALT`.

But I guess it is true that:

1. that goes away if we move back in the direction of direct RV32 execution; and
2. most users would think in terms of the semantics of the high-level program and not the semantics of the compiled assembly, in which case even the NVM behavior could still be "unexpected".

So yeah, I guess for the moment I withdraw this, but it is something we should be thoughtful around being defensive about going forward (maybe at the level of program hash checking).

jolt/src/lib.rs
mx00s: +1, extracting to a named constant can also help keep it self-documenting.

slumber:
Used fork is now at https://github.com/nexus-xyz/jolt so anyone can modify it

My proposed solution to missing ecalls was to update the fork to check registers read according to our convention -- (imo) the simplest and most straightforward thing to do.
Since we couldn't agree on it and the decision was made to pull me from working on the VM code -- I left it as is and the default prover is `nova-seq`, as it used to be.

@sjudson feel free to push any changes to this/fork branches.

sjudson (approved):

PR #164, opened 2024-05-28
Convert NVM Back to RV32 Base

sjudson:
In a line of work culminating in https://github.com/nexus-xyz/nexus-zkvm/pull/53, we introduced a dedicated Nexus VM (NVM) ISA. Although this ISA is similar to RISC-V32i, it is not identical, and further uses a single 64-bit encoding for all instructions.

This change had the benefit of producing slightly smaller circuits (due to the circuit simplifications enabled by the single instruction encoding), as well as enabling some additional simple optimizations and other changes that had some forward-looking potential benefits as the NVM develops, aimed at the unique nature of a zkVM. On the other hand, this change also introduced a potentially unsound transformation from RISC-V to the NVM ISA that we would have to maintain, and also meant that existing tooling infrastructure and other zkVM techniques designed for RV32i would not be directly usable on the code the zkVM was actually executing/proving.

Recent developments in the zkVM space, particularly the significant performance improvements enabled by lookup arguments, have tipped the balance back in the other direction, and mean that it was probably a bit premature to move away from direct RV32i execution/proving. So this PR "reverts" that line of work, while also backporting new functionalities added since (e.g., private input tape reading and an unchecked, paged memory model).

This PR also formalizes an idea implicit in our existing RV32 implementation, which is to introduce an RV32Nexus ISA extension that captures overridden or new instructions (especially precompiles). At the moment, this extension includes overloaded versions of the `ECALL` and `EBREAK` instructions, which are modified to the include a `rd`component (porting https://github.com/nexus-xyz/nexus-zkvm/pull/150), thereby simplifying circuit generation. The VM now also keeps track of which instruction sets (RV32i or RV32Nexus) are used during the course of a given execution.

We can think of the NVM ISA as now being composed of the RV32i + RV32Nexus ISAs, as well as any future extensions we might develop or import (such as multiplication or floating-point extensions).

The following snippet captures how this is defined in the code:

```rust
    // BEGIN RV32Nexus EXTENSION

    // ECALL: An overload of the RV32i ECALL instruction, with explicit return.
    //
    //        In the RV32i spec, ECALL looks like:
    //
    //           00000 00 00000 00000 000 00000 11100 11
    //
    //        We overload the instruction as:
    //
    //           00000 00 00000 00000 000 {-rd} 11100 11
    //
    //        We then use rd as the return location for the ecall. By making the return
    //        explicit in this way, circuit generation is much cleaner since the memory
    //        updates involved are all captured explicitly in the instruction.
    ECALL { rd: u32 },

    // EBREAK: An overload of the RV32i EBREAK instruction, with explicit return.
    //
    //        In the RV32i spec, EBREAK looks like:
    //
    //           00000 00 00001 00000 000 00000 11100 11
    //
    //        We overload the instruction as:
    //
    //           00000 00 00001 00000 000 {-rd} 11100 11
    //
    //        We then use rd as the return location for the ebreak. By making the return
    //        explicit in this way, circuit generation is much cleaner since the memory
    //        updates involved are all captured explicitly in the instruction.
    EBREAK { rd: u32 },
```


vm/src/circuit/r1cs.rs
```diff
@@ -113,6 +108,7 @@ impl R1CS {
         if self.vars.contains_key(name) {
             panic!("local variable override {name}");
         }
+        //debug_assert!(!self.vars.contains_key(name));
```
mx00s, line 111: Should this be removed or uncommented?

vm/src/circuit/r1cs.rs
```diff
@@ -190,24 +174,8 @@ impl R1CS {
...
-    }
-
-    pub fn set_not(&mut self, output: &str, input: &str) -> usize {
-        self.set_field_var(output, ONE - self.get_var(input));
-        self.not(output, input)
-    }
-
-    pub fn to_bits(&mut self, name: &str, val: u32) -> usize {
+    #[allow(clippy::wrong_self_convention)]
+    pub fn to_bits(&mut self, name: &str, val: u32) {
```
mx00s, line 178: This used to return a `usize`, specifically the `vj` value.

I admittedly don't follow this part yet, but am curious: is the output perhaps still needed or do the side-effects of calling this mutable method suffice?

vm/src/error.rs
```diff
@@ -3,17 +3,37 @@ use thiserror::Error;
...
-    UnknownSyscall(u32, u32),
+    #[error("unknown ecall {1} at pc:{0:x}")]
+    UnknownECall(u32, u32),
 
-    /// Invalid memory address
-    #[error("invalid memory access {0:x}")]
-    SegFault(u32),
+    /// An error occurred reading file system
+    #[error(transparent)]
+    IOError(#[from] std::io::Error),
```
mx00s, line 24: nit: I realize this wasn't actually a change, but the comment seems overly prescriptive and susceptible to going stale (if it hasn't already).

```suggestion
    /// An I/O error occurred
    #[error(transparent)]
    IOError(#[from] std::io::Error),
```

vm/src/eval.rs
```diff
@@ -1,27 +1,32 @@
...
+    /// ISA registers
+    pub regs: Regs,
     /// Syscall implementation
     pub syscalls: Syscalls,
+    /// current instruction
+    pub inst: Inst,
+    /// internal result register
+    pub Z: u32,
+    /// used instruction sets
+    pub instruction_sets: HashSet<InstructionSet>,
```
mx00s, line 27: nit: Is this needed now? I see where it's initialized in the `new()` constructor and inserted into in `eval_inst()`, but haven't yet found any reads.

I'm not opposed to including it if it's going to help, but also wonder if we may find a more robust abstraction later. In this context, the `Inst` used for the `inst` already specifies a sort of instruction set that the `NexusVM` should be able to handle.

vm/src/rv32/parse.rs
```diff
@@ -210,8 +210,8 @@ pub(crate) fn parse_u32(word: u32) -> Option<RV32> {
         OPC_FENCE => FENCE,
 
         _ => match word {
-            0x00000073 => ECALL,
-            0x00100073 => EBREAK,
+            0x00000573 => ECALL { rd: rd(word) },
+            0x00100573 => EBREAK { rd: rd(word) },
```
mx00s, line 214: I'm still digesting the code, so bear with me. I follow how this decodes the `RV32Nexus` forms of `ECALL` and `EBREAK` when `rd` is `10` (decimal), but is the intention to support all possible values of `rd` encoded in bits 7-11 (inclusive)?

I suppose as long as the only supported inputs are rv32i-encoded instructions and we control the mapping (`translate_nexus()` below), then we're the only direct user of this RV32Nexus ISA.

Assuming fixing `rd := 10` is sufficient for our needs, what do you think about either ...

1. Restricting the RV32Nexus EXTENSION to only support `rd` encoded as `10` for now or
2. Reverting to `ECALL` and `EBREAK` variants not having any associated data and merely changing their semantics in RV32Nexus EXTENSION to change where/how they return.

sjudson: Good call, will generalize, even if the VM will always use `rd = 10` for now.

vm/src/circuit/r1cs.rs
sjudson: Side-effects suffice, as the constraint system modifies itself.

vm/src/circuit/r1cs.rs
sjudson: I think we can remove this, will do.

vm/src/eval.rs
sjudson: This will get used as soon as I rebase on https://github.com/nexus-xyz/nexus-zkvm/pull/159, in order to block ecalls with Jolt until I've added support for it.

vm/src/rv32/parse.rs
```diff
@@ -241,6 +241,15 @@ fn inst_size(b0: u8, b1: u8) -> u32 {
     }
 }
 
+/// translate RV32i instructions to RV32Nexus instructions
+fn translate_nexus(word: u32) -> u32 {
+    match word {
+        // ecall     // ebreak
+        0x00000073 | 0x00100073 => word | (0b101 << 8), // set rd = 10
```
mx00s, line 248: nit: The relationship between 10 and the instruction encoding of `rd` in bits 7-11 would be clearer like this, at least to me.

```suggestion
        0x00000073 | 0x00100073 => word | (0b1010 << 7), // set rd = 10
```

mx00s (changes requested):
LGTM after the `rd` for `ECALL` and `EBREAK` is generalized to support all possible 32 registers!

vm/src/rv32/parse.rs
sjudson: Should be resolved now.

jolt/src/parse.rs
```diff
@@ -82,13 +84,16 @@ fn parse_instructions(elf: &ElfBytes<LittleEndian>, data: &[u8]) -> Result<Vec<I
...
-                );
-                return Err(Error::Unsupported(inst.inst));
+            match inst.inst {
+                RV32::ECALL { .. } | RV32::EBREAK { .. } => {
+                    tracing::debug!(
+                        target: LOG_TARGET,
+                        ?addr,
+                        "Unsupported instruction",
+                    );
+                    return Err(Error::Unsupported(inst.inst));
```
mx00s, line 94: Based on a discussion with @sjudson, I anticipated the Jolt prover wouldn't fully work yet, at least not with `ECALL`s, but I also wanted to test and document the latest behavior.

I tried out the Jolt prover using the steps described at https://github.com/nexus-xyz/nexus-zkvm/pull/159#issuecomment-2120903704, except with `--impl jolt` since it's no longer the default, and on the HEAD of this branch instead, currently adc3886.

The default "Hello, World" program fails with the following error:

```shell
$ cargo nexus prove --impl jolt
   Compiling proc-macro2 v1.0.84
   Compiling unicode-ident v1.0.12
   Compiling syn v1.0.109
   Compiling nexus-rt v1.0.0 (.../nexus-zkvm/only-riscv/runtime)
   Compiling quote v1.0.36
   Compiling nexus-rt-macros v0.1.0 (.../nexus-zkvm/only-riscv/runtime/macros)
   Compiling pr-164 v0.1.0 (/private/tmp/pr-164)
    Finished release-unoptimized [unoptimized] target(s) in 3.03s
Error: Instruction isn't supported: ecall x10
```

Based on the error message, it must be coming from the highlighted line. Again, this isn't surprising and on its own doesn't necessarily warrant blocking the PR since Jolt isn't the default prover.

mx00s:
While testing I discovered a regression with the current default prover implementation, `nova-seq`.

First, reinstall the `cargo-nexus` binary from this branch:

```shell
$ git rev-parse --short HEAD
adc3886

$ cargo install --path tools
```

Generate a new Nexus project and point the runtime dependency to the local clone of this branch.

```shell
$ cd /tmp
$ cargo nexus new pr-164
$ cd pr-164
$ cargo add nexus-rt --path <<LOCAL_CLONE>>/runtime
```

And finally, run the prover followed by the verifier:

```shell
$ cargo nexus prove
   Compiling proc-macro2 v1.0.84
   Compiling unicode-ident v1.0.12
   Compiling syn v1.0.109
   Compiling nexus-rt v1.0.0 (.../nexus-zkvm/only-riscv/runtime)
   Compiling quote v1.0.36
   Compiling nexus-rt-macros v0.1.0 (.../nexus-zkvm/only-riscv/runtime/macros)
   Compiling pr-164 v0.1.0 (/private/tmp/pr-164)
    Finished release-unoptimized [unoptimized] target(s) in 2.32s
  Setting up public parameters for IVC ... 16.7s
    Finished in 16.7s
Hello, World!
  Loading public parameters ... 4.1s
 Finished in 4.1s
  Computing step 0 ... 653ms
  Computing step 1 ... 744ms
  Computing step 2 ... 904ms
  Computing step 3 ... 944ms
  Computing step 4 ... 929ms
  Computing step 5 ... 926ms
  Computing step 6 ... 948ms
     Proved 7 step(s) in 6.1s; 18.51 instructions / second
  Saving proof ... 23ms
Finished in 23ms

$ cargo nexus verify
  Loading public parameters ... 4.5s
 Finished in 4.5s
  Verifying proof .   0ms
  2024-05-29T18:21:13.403012-07:00 ERROR nexus-tools: Proof is invalid, err: not satisfied, k: 16, nova_impl: nova-seq
```

When I repeated the following manual test on `main` (rev bc517fe) I found it succeeds; the regression must be in this branch.

jolt/src/parse.rs
mx00s: When I removed any source of `ECALL`s in the program to prove/verify, the `jolt` impl worked as expected. I've not audited to see what, if anything, the optimizer elided as dead code, but it's good to see there's no apparent regression here.

```diff
diff --git a/src/main.rs b/src/main.rs
index 69a7814..f271880 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -5,5 +5,8 @@ use nexus_rt::write_log;
 
 #[nexus_rt::main]
 fn main() {
-    write_log("Hello, World!\n");
+    let mut i = 0;
+    while i < 10 {
+        i += 1;
+    }
 }
```

Prove then verify with `--impl jolt`:

```
$ cargo nexus prove --impl jolt
   Compiling proc-macro2 v1.0.84
   Compiling unicode-ident v1.0.12
   Compiling syn v1.0.109
   Compiling nexus-rt v1.0.0 (.../nexus-zkvm/only-riscv/runtime)
   Compiling quote v1.0.36
   Compiling nexus-rt-macros v0.1.0 (.../nexus-zkvm/only-riscv/runtime/macros)
   Compiling pr-164 v0.1.0 (/private/tmp/pr-164)
warning: unused import: `nexus_rt::write_log`
...
    Finished release-unoptimized [unoptimized] target(s) in 2.64s
  Preprocessing bytecode ... 533ms
       Finished in 533ms; bytecode size: 55
Executing program...
Executed 104 instructions in 8.011541ms
  Proving program execution ... 2.5s
 Finished in 2.5s
  Saving proof ... 1ms
Finished in 1ms

$ cargo nexus verify --impl jolt
  Preprocessing bytecode ... 480ms
       Finished in 480ms; bytecode size: 55
  Verifying proof ... 143ms
   Finished in 143ms
```

sjudson:
Appears to be an issue with tracking branching instructions for larger `k`, just pushed up some code that seems to solve it on my end.

mx00s:
I repeated the same test with the default prover on f60ef84. It worked!

mx00s (approved):

Issue #169, opened 2024-06-06
Enable Curve + Commitment Configuration


sjudson:
At present, our proving scheme/compression implementations are written over generics. However, our vm is written over a specific curve pair [here](https://github.com/nexus-xyz/nexus-zkvm/blob/7fa6fe89ada8370f4a7f13cbaf96fb67e012dac1/vm/src/circuit/r1cs.rs#L16) and [here](https://github.com/nexus-xyz/nexus-zkvm/blob/7fa6fe89ada8370f4a7f13cbaf96fb67e012dac1/vm/src/memory/cacheline.rs#L5), while the `prover` crate (soon to be moved under the `api` crate) instantiates a particular set of choices for these generics:

- primary curve: `bn254`
- secondary curve: `grumpkin`
- commitment scheme: `PedersenCommitment`  
- polynomial commitment scheme: `Zeromorph`
- random oracle: `Poseidon`

In the future, the SDK will want to support user-configuration of these objects (with sensible defaults). So we should replace these concrete interfaces with generic ones all the way through, so that the SDK is not constrained to these particular choices.

slumber:
I think it is up to discussion whether the end consumer crates should be generic.

sjudson:
That's ultimately a UI/UX decision though. So though we may choose to limit how much configuration is exposed to the end-user, the `vm` arithmetization certainly shouldn't be curve-specific, while the `prover` crate needs to be uncoupled from the `tools` in order to feed both it + and the `sdk`, and in my opinion shouldn't overly constrain the options it presents to those interfaces.

mx00s:
This sounds sensible to me.

Regarding whatever opinionated sets of generic arguments we pass in production contexts, whether the generic parameters exposed to users directly or not, it may be helpful to consider, "if I were to ask for a security audit, how would I specify the scope?"

mx00s:
We may also find it makes sense to limit the number of generic parameters. Each additional generic parameter has a multiplicative effect on the set of generic type arguments to consider, and that can get complicated.

An alternative is to make a trait with associated types. You can then pass around sets of supported compositions of types as a single generic argument which is an implementation of that trait. So, to illustrate, something like:

```rust
trait Configuration {
    type PrimaryCurve;
    type SecondaryCurve;
    type CommitmentScheme;
    type PolynomialCommitmentScheme;
    type RandomOracle;

    ...
}

struct X;

impl Configuration for X {
    type PrimaryCurve = bn254;
[...]

Issue #165, opened 2024-05-30
Add end-to-end zkVM regression test


mx00s:
Detect regressions like https://github.com/nexus-xyz/nexus-zkvm/pull/164#issuecomment-2138511367 automatically in the test suite.

sjudson:
Resolved by rewrite of codebase for the new Nexus 3.0 machine (https://github.com/nexus-xyz/nexus-zkvm/pull/365).

Issue #148, opened 2024-04-22
Prover SDK tracking issue


slumber:
The main prover functionality is gathered in `nexus-prover` crate, which does all sort of stuff: generate public parameters, print nice CLI output, compress proofs etc.

I propose to remove it in favor of `nexus-sdk` which provides abstracted interface like
```rust
pub trait Prover {
    /// Output type.
    type Proof;
    
    /// Public parameters needed for generating proofs, if any.
    type PublicParams;
    
    fn prove(&self, params: &Self::PublicParams) -> Self::Proof;
    
    /// or something like prove_step to facilitate usage of "stepped" cli output.
}
```

And then use it
```rust
pub struct NexusProver<P: Prover = CompressedNova> {
    /* private */
}

pub struct NexusVM { }
```

This SDK is then integrated into `nexus-tools` to be reused with CLI and pretty output.

slumber:
@govereau @sjudson @danielmarinq

slumber:
> You propose doing this alongside or in place of the `nexus-api` approach (#146) @slumber? For dev I think it'll be important to expose VM interfaces alongside the prover interface so that users can develop without needing to invoke the proof apis.

`nexus-api` doesn't look like proper sdk and is more of a short-term solution, as it just re-exports what we currently have.
This issue is about generically wrapping the proving system and the vm into an extendable interface which we ourselves can reuse.

sjudson:
I'm amenable to really taking the time to think through the api design in a more comprehensive way than the https://github.com/nexus-xyz/nexus-zkvm/pull/146 approach, which is really an attempt to get programmatic usage "out the door" in an at least somewhat constrained and opinionated way.

I do think it makes sense though to have an independent public facing API from the prover API that is also consumed internally though.

slumber:
> I'm amenable to really taking the time to think through the api design in a more comprehensive way than the #146 approach, which is really an attempt to get programmatic usage "out the door" in an at least somewhat constrained and opinionated way.
> 
> I do think it makes sense though to have an independent public facing API from the prover API that is also consumed internally though.

This is relevant because we already have Nova(seq), Nova(pcd), Nova(pcd+spartan), HyperNova soon, maybe Jolt which is going to be different.
There's no convenient way of switching between these. We could also add things like curve choice or commitment scheme later.

sjudson:
[...]

PR #170, opened 2024-06-07
Move Towards Generic API

sjudson:
At present, we have a variety of pipelines for each `frontend` -> `vm` -> `proving scheme` implementation.

- `tools` (minus `src/command/jolt.rs`) -> `prover` -> `nova`
- `prover` -> `nova`
- `tools/src/command/jolt.rs` -> `jolt`
- `api` -> `prover` -> `nova` 

where the last has to take care to avoid a lot of coupling between the CLI and the `prover` crate. These pipelines are often independent from each other when they should be coherent, or intertwined with each other when they should be separated. This is not so big a deal at present, but in the short-term make it difficult to cleanly implement a future-proofed (as much as possible) SDK, and in the long-term will get unwieldy over time as various frontends and backends proliferate.

This PR is an initial move to a cleaner approach, where the `api` crate concentrates various backends that can be consumed by the various frontends (for now the CLI, Rust SDK, and `network` crate, in the future more network functionality as well as SDKs for other languages and environments).

In practice, this PR does three major things:

1. It moves the `config` and `prover` crates under the `api` crate, developing a single interface for frontends to access.
2. It slims down the `prover` crate, moving all CLI-dependent functionality into the `tools` crate and removing the vestigal SDK that was implemented within it.
3. It passes the Jolt implementation through the `api/prover` crate as well, again with the goal of having a single interface for frontends to access.

Also relevant is https://github.com/nexus-xyz/nexus-zkvm/issues/169. At present, the API crate is still "opinionated" about configuration of generic parameters, e.g., choices of curve pairings for Nova, random oracles, and commitment schemes. In the future we may want to let the various frontends to configure these generic parameters, which would require making the code in `api` suitably generic for the frontends to consume as needed.


config/Cargo.toml
```diff
@@ -1,24 +0,0 @@
...
-version = { workspace = true }
-edition = { workspace = true }
-homepage = { workspace = true }
-repository = { workspace = true }
-publish = { workspace = true }
-keywords = { workspace = true }
-categories = { workspace = true }
-
-[dependencies]
-serde = { version = "1.0", features = ["derive"] }
```
slumber, line 13: config crate is used for various workspace configurations, and should be minimal.

An example is I want to add configs for RPC, like disconnect timeout, connections limit etc. With you PR I need to depend on `api` which bring, for example, spartan.

Work towards sdk, as I imagine it, should 
1. introduce `nexus-sdk` crate with generic trait interfaces
2. migrate workspace to using it

api/examples/prover_run.rs
```diff
@@ -40,7 +40,14 @@ fn main() {
...
     println!("Proving execution of length {}...", trace.blocks.len());
-    let proof = prover::prove::prove_seq(&public_params, trace).expect("error proving execution");
+    let tr = prover::nova::init_circuit_trace(trace).expect("error constructing circuit");
+
+    let mut proof =
+        prover::nova::prove_seq_step(None, &public_params, &tr).expect("error proving execution");
+    for _ in 1..tr.steps() {
+        proof = prover::nova::prove_seq_step(Some(proof), &public_params, &tr)
+            .expect("error proving execution");
+    }
```
mx00s, line 50: Is there perhaps an off-by-one error here?

Since `..` is exclusive at the end, this executes one time less than the number of blocks in the trace. If we want one-call per block you could change the iterator's start value from `1` to `0`. Alternatively, there's `..=` which is inclusive at the end.

api/src/prover/nova/circuit.rs
```diff
@@ -15,16 +15,16 @@ pub use ark_relations::{
...
     trace::{trace, Trace},
 };
 
-use crate::error::*;
-use crate::types::*;
+use super::error::*;
+use super::types::*;
 
-pub struct Tr(pub Trace<Path>);
+pub struct Tr<M: Memory>(pub Trace<M::Proof>);
```
mx00s, line 25: 👍

tools/src/command/prove.rs
```diff
@@ -114,55 +114,181 @@ fn local_prove(
...
+                        break;
+                    }
+
+                    vs = vs
+                        .chunks(2)
+                        .map(|ab| {
+                            let _guard = term_ctx.display_step();
+                            let c = nexus_api::prover::nova::prove_par_parent_step(
+                                &state, &tr, &ab[0], &ab[1],
+                            )?;
```
mx00s, line 207: Can `vs.len()` ever be odd here? If so, the `ab[1]` indexing will sometimes panic because [slice's `chunks` method](https://doc.rust-lang.org/std/primitive.slice.html#method.chunks) the last chunk returned may be shorter than the specified length when there's a remainder.

api/examples/prover_run.rs
sjudson: The initial invocation at L45 handles the first step.

This function call should not be user facing (it'll be hidden behind the SDK and the CLI), but thinking about it it's probably worth introducing a more misuse resistant variant for when we don't need to hook in telemetry (like here), so I'll add that.

tools/src/command/prove.rs
sjudson: Ah good call. Used correctly it always is, this is another artifact of losing misuse-resistance in order to move to a step-by-step interface to unhook the telemetry. I'll add an additional check.

config/Cargo.toml
sjudson: I'm amenable to moving the `config` crate back out, but at that point would say I think it makes more sense to move the configuration logic to the various crates its components refer to, rather than to have this small crate dangling off to the side from the `api` also trying to combine the various functionalities spread around into one coherent interface. But if you feel strongly about keep `config` where and how it is I can move it back. 

As for the proposed flow, I agree in principle. The problem in practice is that the `prover` crate is so tied into the CLI at present that I can't make a reasonable, generic SDK that uses it in the way I'd like without doing this restructuring first.

sjudson:
Restored the original `config` crate at @slumber's request.

api/Cargo.toml
```diff
@@ -10,6 +10,53 @@ keywords = { workspace = true }
 categories = { workspace = true }
```
slumber, line 10: This file needs a cleanup, e.g. I don't think it should depend on `tui` and have `verbose` feature anymore

slumber:
As for the sdk shape: I imagined we could start with very minimal 
```rust
// call it nexus-prover-core crate or smth
trait Prover {}
```
which should be as lightweight as possible

Then, all prover crates would depend on it.
```rust
// nexus-nova
impl Prover for NovaProver {}
// nexus-jolt
impl Prover for JoltProver {}
```
Finally,
```rust
// nexus-sdk
pub use Prover;
#[cfg(feature = "nova")]
pub use nexus_nova::NovaProver;
#[cfg(feature = "jolt")]
pub use nexus_jolt::JoltProver;
```

Maybe not so much configurable in terms of curves/fields/k/etc., perhaps snips some prover-specific features, but clean and easy to use.

sjudson:
Converted back to draft to enable some other relevant code to come through first, that this will be adapted to.

mx00s:
> Thoughts on this @mx00s?

To start, I'm going to avoid using the terms API and SDK.

I agree with the idea of leveraging clearly defined interfaces to improve cross-team coordination.

> I've worked in a context before where you have a proliferation of frontend SDKs/CLI tooling and backends, and the only reason it was manageable from a cross-team coordination standpoint was because the web API that linked everything together provided a single, concrete interface that enabled clear interoperability.

That will be necessary as 

1. we iteratively improve implementations of those interfaces,
2. create new implementations, and
3. support a growing variety of consumers.

I also appreciate the benefits of designing simple interfaces starting top-down as opposed to bottom-up as @slumber seems to suggest:

> going back to review https://github.com/nexus-xyz/nexus-zkvm/pull/170#discussion_r1633268126, I don't understand the approach of rewriting anything before the actual sdk is implemented. Meaning: come up with generic trait interface (https://github.com/nexus-xyz/nexus-zkvm/issues/148 original description seems to be ignored), introduce it as a crate into the workspace, only then start migrating code like tools or vm to use these interfaces.

In terms of API and SDK dichotomy as it's been presented so far, my understanding is the API is meant to be a lower-level abstraction for IVC and the SDK will eventually depend on it, providing higher-level interfaces that encapsulate more of the details. I can imagine this working well, but admittedly am hazy on who the intended consumers of the SDK would be and what details relevant to the API they'd benefit from having encapsulated.

It's abundantly clear we need clean, simple interfaces to support the aforementioned coordination points over time. Whether to design such interfaces working bottom-up or top-down is somewhat a matter of taste, but I am admittedly biased toward the top-down approach to optimize for ergonomic UX. Admittedly, sometimes the top-down approach means discovering while wiring up the internals that the high-level interfaces must be adjusted, but that's okay.

I've mentioned it elsewhere, but if we align on the top-down approach I think we'd benefit from:

1. implementing end-to-end tests https://github.com/nexus-xyz/nexus-zkvm/issues/165
2. defining high-level interfaces that make those tests clean
3. promoting those high-level interfaces to a public crate that serves as a coordination point, whatever we choose to call it

By the way, like @sjudson I don't see "api" as usually pertaining to networking. I suppose OpenAPI/Swagger may have shifted norms in that direction a bit.

sjudson:
Removing from draft due to merge of https://github.com/nexus-xyz/nexus-zkvm/pull/185

config/Cargo.toml
```diff
@@ -21,4 +21,4 @@ url = { version = "2.5.0", features = ["serde"] }
 clap = { workspace = true, optional = true } 
 
 [features]
-clap_derive = ["clap"]
+clap_derive = ["clap"]
```
slumber, line 24: it's best to checkout `/config/*` to match main

slumber: My comment is about the whole directory, notice there're toml files recommitted as well as a backup mod.rs

sjudson: Ugh, I did checkout the whole directory, not sure why git didn't clear deleted files...

network/rpc/server/Cargo.toml
```diff
@@ -21,9 +21,6 @@ path = "src/bin/client.rs"
 [dependencies]
 nexus-rpc-traits = { path = "../traits", features = ["server"] }
 nexus-rpc-common = { path = "../common" }
-nexus-prover = { path = "../../../prover" }
-nexus-vm = { path = "../../../vm" }
-nexus-config = { path = "../../../config" }
```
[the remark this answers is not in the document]
slumber: Notice how bringing `nexus-api` here also requires compiling jolt as a dependency, while it's unused.

Logically one would feature-gate api modules, but then either all of these features get propagated into depending crate, or disabled by default, making development experience worse.

tools/src/lib.rs
```diff
@@ -43,3 +43,8 @@ const ENV: &str = r#"
 NEXUS_VM_K=16
 NEXUS_VM_PROVER=nova-seq
 "#;
+
+#[cfg(feature = "verbose")]
+pub(crate) const TERMINAL_MODE: nexus_tui::Mode = nexus_tui::Mode::Enabled;
+#[cfg(not(feature = "verbose"))]
+pub(crate) const TERMINAL_MODE: nexus_tui::Mode = nexus_tui::Mode::Disabled;
```
slumber, line 50: "verbose" feature is not defined as the cli is always verbose. `tools` use `TerminalHandle::new_enabled()`

jolt/src/lib.rs
```diff
@@ -59,43 +52,7 @@ impl<M: Memory> VM<M> {
     }
 }
 
-pub fn preprocess<M: Memory>(vm: &VM<M>) -> JoltPreprocessing {
```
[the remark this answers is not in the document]
sjudson: As discussed, goal here is to have a single interface that re-exports necessary functionality for various frontends/network components to rely upon.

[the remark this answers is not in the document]
slumber: I get that api re-exports, but this isn't re-exporting, this is moving jolts logic into a different crate, scattering it over the workspace.

Meaning, if I look into `nexus-jolt` code, there's code for translating trace, but no code for generating actual proof. So even if I want to use it bypassing api, I can't.

All jolt methods are public, why is the following not sufficient?
```rust
// nexus-api
pub use nexus_jolt::*;
```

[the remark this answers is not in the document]
sjudson: I see, yes, I can move in that direction.

network/rpc/server/Cargo.toml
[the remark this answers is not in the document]
sjudson: Fixed.

tools/src/lib.rs
sjudson: Fixed.

jolt/src/lib.rs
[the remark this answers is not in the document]
sjudson: Fixed.

slumber:
I assume https://github.com/nexus-xyz/nexus-zkvm/pull/170#discussion_r1645049397 is not going to be addressed. 

Given the amount of design-discussions about _potential issues_ we will face in years after implementing SDK, and that this PR already makes it worse to build on top of workspace crates, I'll again outline how it just doesn't work.

My impression is it's crafted to facilitate `tools` which never needed it.

Please get an approval from @mx00s to unblock the merge.

sjudson:
After some further discussion, added feature gating @mx00s @slumber: https://github.com/nexus-xyz/nexus-zkvm/pull/170/commits/fde39c00a8df8ac85d0a1df36eafeda9da834b60

sjudson:
Updated dependency graph:

image

replacing 

image

mx00s:
Nice. For reference, the following command can be used to produce those dependency graphs:

```bash
cargo install cargo-depgraph
cargo depgraph --workspace-only | dot -Tpng > graph.png
```

mx00s (approved):
With the dependency graph reconfigured, we're better positioned to iterate on `nexus-api` and support the many-to-many dependence between prover frontends and backends.

Thanks for adding the prover feature flags too. This should help as continue implementing/improving provers and even vary defaults per deployment environment, for instance.

LGTM!

PR #172, opened 2024-06-11
Fix example code in doc

yugocabrio:
It seemed to have forgotten to update the Fibonacci code only in the doc.


danielmarinq:
Thanks @yugocabrio! @slumber can you help with this?

sjudson:
Thanks for this, approved and merged.

PR #181, opened 2024-06-12
Update README.md

Budalebah:
added command to run the project in the right directory

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,10 @@ fn main() {
 ### 3. Run your program
 
 ```bash
+cd nexus-project
+```
+
+```bash
 cargo nexus run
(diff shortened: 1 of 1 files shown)


Budalebah (commented):
added images file and images for pathways of svg.mdx files

sjudson:
I'm not sure why the images are included, but they don't serve any purpose within the repo.

Otherwise I think the readme is fine as it is: we assume the reader is a developer familiar with the basics of `cargo` and able to navigate the resultant project directory.

Budalebah:
yes but the current links are empty and i filled them. for cd , not all testers are professional so added it for newbies, but you know

PR #182, opened 2024-06-13
chore: remove  License section

suiyuan1314:
As the software is licensed, so i removed  the word "not" to avoid misunderstanding.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -16,4 +16,4 @@ npm run dev
 
 ## License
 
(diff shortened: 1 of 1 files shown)


sjudson (changes requested):
It would probably be best to just delete that entire License section, as it's been superseded from when it was written.

If you'd like to update the PR to fully remove it we'll be happy to merge.

suiyuan1314:
done @sjudson

sjudson (approved):
Thanks!

PR #184, opened 2024-06-13
Fix Bug Report Template

sjudson:
Typo leading to error that prevents the template from working.


sjudson:
Now working, can be seen at: https://github.com/nexus-xyz/nexus-zkvm/blob/fix-templates/.github/ISSUE_TEMPLATE/bug_report.yml

slumber (approved):

Issue #187, opened 2024-06-17
remove `tools-dev`


slumber:
Initially it was introduced to facilitate logging, config management and launching network.

Both `network` and `prover` crates are going to be removed soon. The only functionality left is parsing toml configs into env file.

Keeping it inside `config` crate is an option, however while revisiting it, I think doing
```rust
let config = SomeProverConfig::from_env()
    .unwrap_or_default(); // Where default() returns configs for tests.
                          // unwrap_or_else(test_config) is better, leaves a INFO log about test config used
```
Is simpler than tracking toml files being in sync with Rust structs.

PR #185, opened 2024-06-13
jsonrpc-based prover server

slumber:

PR #191, opened 2024-06-18
remove tools-dev crate

slumber:
Resolves https://github.com/nexus-xyz/nexus-zkvm/issues/187

See the issue description. Initially, I introduced `tools-dev` for making it easier to run `network` nodes with different configurations. I imagined it will be used for running local integration tests.

Potentially, we could extend toml configurations and compile everything into a single `env` file used locally.

Finally, tools-dev doesn't require recompiling to run other crates, as it internally invokes workspace cli s.

In practice, dev cli had fallen behind the actual cli and only holds definitions for commands. `nexus-prover` cli is going to be removed, and `network` crate is still broken because of the wrong padding in PCD implementation.

The only functionality left is toml configurations -- this PR removes it as well, see config doc comments. If local config is needed anywhere (for example, RPC), it does:
```rust
let config = SomeConfig::from_env()
    .unwrap_or_else(|error| {
        let config = Default::default();
        tracing::info!(
            target: "some_crate",
            %error,           
            "Using default configuration",
        );
        config
}
``` 
Or `unwrap_or_default`.

Also bumps jolt rev as I ran into a compile error with new rustc.


tools/src/command/new.rs
```diff
@@ -1 +1,77 @@
-pub use nexus_tools_dev::command::common::new::handle_command; // the implementation is shared.
```
slumber, line 1: No changes done to this file aside from `examples_dir`

sjudson (approved):
LGTM, and will make navigating the `tools` code much more intuitive. Also looking at https://github.com/nexus-xyz/nexus-zkvm/pull/185 the new config approach seems more than clean enough for now, and we can make it more robust in the future if need be.

jolt/Cargo.toml
```diff
@@ -14,5 +14,5 @@ strum = "0.25.0"
 ark-bn254 = "0.4"
 
 # latest commit on slumber-arkworks-compat
-jolt-core = { git = "https://github.com/nexus-xyz/jolt", rev = "dfb9f27" }
-jolt-common = { git = "https://github.com/nexus-xyz/jolt", rev = "dfb9f27", package = "common" }
+jolt-core = { git = "https://github.com/nexus-xyz/jolt", rev = "f0f1f12" }
```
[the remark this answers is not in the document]
slumber: The problem is I actually encountered this error while merging two crates. Don't even know what caused it: `cargo check` on jolt works, but installing tools doesn't (probably install always ignores lock file)

The fact that `jolt-core` depends on tokio that caused trouble is just a bad design. Next gen zkVM waiting room.

Agree with the comment, but I don't put much belief into this jolt wrapper.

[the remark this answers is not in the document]
sjudson: Oh weird, okay that makes more sense the.

tools/src/command/public_params.rs
```diff
@@ -141,3 +140,12 @@ pub fn sample_test_srs(args: SRSSetupArgs) -> anyhow::Result<PathBuf> {
     gen_test_srs_to_file(num_vars, path_str)?;
     Ok(path)
 }
+
+// TODO: make it accessible to all crates.
```
mx00s, line 144: nit: This appears to be stale; consider removing.

```suggestion
```

To verify, I ran `cargo doc --open` from the toplevel directory of this branch and saw that `nexus_tools::command::public_params::format_params_file` is public.

mx00s (approved):
LGTM!

As usual, my nits are non-blocking and can be addressed in future PRs when it makes sense.

tools/src/command/public_params.rs
slumber: this TODO means this function should be somewhere else. Surely you don't want to depend on `nexus-tools` just to be able to format a parameters filename. Maybe it should be part of nova, config, etc. :man_shrugging:

PR #195, opened 2024-06-19
nexus-rt overhaul

### LINKED ISSUE #167: Rework nexus-runtime macros
Essentially

https://github.com/nexus-xyz/nexus-zkvm/blob/49b89a8a1b43d4d0d7edf2670f4fc3378cf40c39/runtime/macros/src/lib.rs#L21-L22

And

https://github.com/nexus-xyz/nexus-zkvm/blob/49b89a8a1b43d4d0d7edf2670f4fc3378cf40c39/runtime/build.rs#L24-L26

The crate itself and the macro only work for `riscv-*` targets. It worsens developer experience, both users and ours:
1. It's impossible to run any tests on nexus projects.
2. rust-analyzer panics in the build script of the nexus-runtime.

The macro should be a no-op when `#[cfg(not(target_arch = "riscv32"))]` as well as the runtime should be simply disabled.

Nexus projects will migrate to
```rust
#[cfg_attr(not(test), no_std)]
#[cfg_attr(target_arch = "riscv32", no_main)] 

// or simpler
#[cfg_attr(target_arch = "riscv32", no_std, no_main)] // std should always be disabled unless used for tests
```

Default `.cargo/config` will get in the way, so also needs changes.

slumber:
Resolves #167 

`cargo run` runs code natively, and `cargo nexus run` runs it on Nexus VM. This PR also refactors `nexus-rt-macros`.


runtime/macros/src/lib.rs
```diff
@@ -1,140 +1,11 @@
...
-
-#![deny(warnings)]
-
-extern crate proc_macro;
-#[macro_use]
-extern crate quote;
-extern crate core;
-extern crate proc_macro2;
-#[macro_use]
-extern crate syn;
```
slumber, line 11: This file is a copy of `riscv-rt` as the comment says, but it doesn't fit our needs and can be reduced to a much more concise implementation.

runtime/macros/src/lib.rs
```diff
@@ -1,140 +1,11 @@
...
+mod entry;
+mod parse_args;
+
 #[proc_macro_attribute]
 pub fn main(args: TokenStream, input: TokenStream) -> TokenStream {
-    let f = parse_macro_input!(input as ItemFn);
-    let a = parse_macro_input!(args with Punctuated::<Meta, syn::Token![,]>::parse_terminated);
-
-    // check the function arguments
-    if f.sig.inputs.len() > 3 {
```
slumber, line 27: All inputs should be forbidden as using them causes a panic.

runtime/macros/src/lib.rs
```diff
@@ -1,140 +1,11 @@
...
-        && f.sig.asyncness.is_none()
-        && f.vis == Visibility::Inherited
-        && f.sig.abi.is_none()
-        && f.sig.generics.params.is_empty()
-        && f.sig.generics.where_clause.is_none()
-        && f.sig.variadic.is_none()
-        && match f.sig.output {
-            ReturnType::Default => true,
-            ReturnType::Type(_, ref ty) => is_simple_type(ty, "u32"),
-        };
```
slumber, line 63: This is excessive. For example, changing `main` visibility is fine, generics would cause an error on their own, empty `where` clause is harmless etc.

On native, users will face an error from original requirements from `fn main()`. On riscv target most of these do nothing.

Only sensible thing to do is to disable arguments and return value.

runtime/src/ecalls.rs
```diff
@@ -1,58 +1,79 @@
...
-    ecall!(2, inp, inp, out);
-
-    if out == u32::MAX {
-        None
-    } else {
-        Some(out.to_le_bytes()[0])
-    } // u32::MAX is used a sentinel value that there is nothing (left) on the input tape
-}
+    /// Read a byte from the private input tape
+    pub fn read_from_private_input() -> Option<u8> {
```
slumber, line 24: This needs to be defined on native, for example always return `None`, read from stdin, etc.

runtime/src/ecalls.rs
```diff
@@ -1,58 +1,79 @@
...
+#[cfg(target_arch = "riscv32")]
 #[macro_export]
 macro_rules! print {
     ($($as:tt)*) => {
-        core::write!(&mut nexus_rt::NexusLog, $($as)*).unwrap();
+        <nexus_rt::NexusLog as core::fmt::Write>::write_fmt(
+            &mut nexus_rt::NexusLog,
+            core::format_args!($($as)*),
+        )
+        .unwrap()
```
slumber, line 58: No longer requires importing `core::fmt::Write`, as supposed to.

runtime/src/ecalls.rs
sjudson: I would vote that this should `panic` on native.

In the future we could explore having some sort of dev infrastructure that hooks into native run code and lets your replicate (private) input type behavior, but for now I think it's better for it to blow up than to do the wrong thing.

README.md
```diff
@@ -72,8 +72,7 @@ This will create a new Rust project directory with the following structure:
 As an example, you can change the content of `./src/main.rs` to:
 
 ```rust
-#![no_std]
-#![no_main]
+#![cfg_attr(target_arch = "riscv32", no_std, no_main)]
```
sjudson, line 75: Just noting for the future that this is an additional motivating to add our own custom target to differentiate vanilla `riscv32` from our NVM-variant that will include, e.g., precompiles

examples/src/bin/memlimit.rs
```diff
@@ -1,15 +1,14 @@
...
 use alloc::vec;
 use core::hint::black_box;
 
-use nexus_rt::{println, Write};
+use nexus_rt::println;
 
 // use 16mb
 // alternatively, remove memlimit argument or otherwise set less than (slightly more than) 6mb and it will fail
-#[nexus_rt::main(memlimit(16))]
+#[nexus_rt::main(memlimit = 16)]
```
sjudson, line 11: This is a nice syntax.

runtime/macros/src/lib.rs
sjudson: Isn't that kinda of the idea of this code though? To affirmatively throw and give the user a clean error message rather than have some code/building silently behave differently than on native?

I guess more generally I'm unclear on why "excessive" is a problem here. It's essentially costless and enforces a clear contract of what code the VM is intended to run. Seems fine to me to leave this in.

sjudson (commented):
Overall this looks like a great start, will make this area of the code much cleaner.

README.md
slumber: I don't follow the motivation, having precompiles doesn't necessarily change the architecture of the target. Besides, CTRL+SHIFT+R "riscv32" to "my_nexus_target".

runtime/src/ecalls.rs
slumber: Doesn't make much difference to me as long as it's documented. Prohibiting using it on native target with panic sounds worse than returning `None`, forcing people to do
```rust
#[cfg(target_arch = "riscv32")]
let input = read_from_private_input();
#[cfg(not(target_arch = "riscv32"))]
let input = read_from_stdin(); // or None
```

runtime/macros/src/lib.rs
slumber: Rust gives you a nice error when you change signature of `main`, but this code doesn't, it's a blind copy-paste.
(note that original descriptive message is propagated by default, unless doing `nexus run`)

My original comments explains the reasoning. If we wan, for example, prohibit usage of generics which would never compile anyway, we should give an explicit error 
```
`main` function is not allowed to have generic parameters
```

etc. We can do it, but I don't see why. Maybe prohibiting `async` is important.

slumber: A note:

An easy way to check expected signature (similar to how Rust does this) is to encode the following into the macro:
```rust
let func = parse_macro_input!(input as syn::ItemFn);
let fn_name = &func.sig.ident;
quote! { 
    #func

    const fn check_main(_: fn()) {}
    const _: () = {
        // Gives a compile error if main is unsafe, has ABI, or any inputs/outputs
        check_main(#fn_name)
    };
}.into()
```

But this isn't desirable until we can optimize this const away with release build

runtime/src/ecalls.rs
sjudson: I agree it's "worse" in some sense, but if we want to match the developer UX we should do it correctly, rather than have a program differ depending on arch.

This is _really important_, since often a program that is proved on the VM might be audited/formally verified off the VM, so not having non-explicit differences in behavior is an issue for soundness.

runtime/macros/src/lib.rs
sjudson: So maybe I misunderstood your original comment, but I took what you wrote to mean that on `riscv` targets you don't necessarily get the nice error message, but instead most of these changes just invisibly stop working.

If you get the nice error message either way then just ignore this comment entirely. I'd just rather explicitly throw a nice error message we write than have different behavior depending on arch.

README.md
sjudson: Oh this was a note for the future, not something that should in anyway change this PR.

Was just noting cause it occurred to me that as the VM starts to differ in behavior from `riscv32` more generally, that we should probably make sure the UX reflects that for explicitness.

runtime/macros/src/lib.rs
slumber: https://github.com/nexus-xyz/nexus-zkvm/pull/195#discussion_r1648860164 the above comment explains how to match macro behavior. At the cost of more steps in debug mode:(

slumber:
@sjudson I agree with the point https://github.com/nexus-xyz/nexus-zkvm/pull/195#discussion_r1648062903

I've made it close to native errors returned on invalid `main` function signature. Also added compile tests, will be useful when we decide to allow `main` arguments or change something else.

sjudson:
@slumber is this ready for full review?

slumber:
> Member



> @slumber is this ready for full review?

it definitely needs some cleanup, but the main idea is implemented.
I pushed the PR as a poc, if we agree on the idea I (or anyone else) can easily finish it.

sjudson:
Okay, well I "approve" of the PoC. Happy to give it a thorough review when it's ready for it @slumber.

examples/src/bin/memlimit.rs
slumber: @sjudson I think while we're at it, two suggestions:
1. Rename `memlimit` to `stack_size`, as I assume this is what this argument configures. Memory size would usually mean heap size, so it creates confusion. E.g. jolt: https://github.com/a16z/jolt/blob/5d1b57cb1dac6b675e90e5a1db1271c44366a6b4/book/src/usage/troubleshooting.md?plain=1#L13
`memory_size` or `heap_size` will be available as a future reserved configuration.

2. Set it in bytes, not megabytes. Otherwise, we're dealing with conversions ourselves. I made it so it saturates, but it used to wrap/panic based on the build type.

runtime/macros/src/entry.rs
```diff
@@ -0,0 +1,53 @@
...
+    }
+    if func.sig.generics.where_clause.is_some() {
+        return Err(Error::new_spanned(
+            &fn_sig.generics.where_clause,
+            "`main` function is not allowed to have a `where` clause",
+        ));
+    }
+
+    Ok(quote! {
+        const _: fn() = main;
```
slumber, line 38: I'm concerned about this. It's correct to do this now, but I don't have a plan about supporting return types on native.
Rust only allows `Result<(), Error>` for main functions. Either we switch from using `main` as a callable function, or introduce some weird syntax like
```rust
nexus_rt::return! { value }; // would print on native and store to memory on riscv, or smth.
```

Maybe it's not that bad?

runtime/src/ecalls.rs
```diff
@@ -1,58 +1,84 @@
...
+    };
+}
+
+#[cfg(not(target_arch = "riscv32"))]
+pub use std::{print, println};
+
+#[cfg(not(target_arch = "riscv32"))]
+pub fn read_from_private_input() -> Option<u8> {
+    panic!("private input is not available outside of NexusVM")
 }
```
slumber, line 84: can be made to return a compile error if we turn it into macro

sjudson: No strong opinion on this, would be easy to migrate given that we could have both in parallel. But this seems fine to me for now.

sjudson (approved):
LGTM. Approving, and happy to re-review if you choose to make some of the outstanding changes mentioned. But otherwise I think it's ready to merge.

examples/src/bin/memlimit.rs
slumber: I decided not to apply it, will file a follow-up. I couldn't make the example work with both array or vec. Setting the limit to 16 **bytes** worked as well.

runtime/macros/src/entry.rs
slumber: One way or another, some magic in macro will be required to make it work on native

PR #197, opened 2024-06-21
Expose HyperNova through API

sjudson:
As implied, exposes our HyperNova implementation through the API to be consumed by the CLI/SDK/network crates.


api/Cargo.toml
```diff
@@ -43,5 +43,6 @@ parallel = [
    "ark-r1cs-std/parallel",
    "nexus-nova/parallel",
 ]
+prover_hypernova = ["dep:nexus-nova", "dep:spartan", "nexus-nova/spartan"]
 prover_nova = ["dep:nexus-nova", "dep:spartan", "nexus-nova/spartan"]
 prover_jolt = ["dep:nexus-jolt"]
```
slumber, line 48: newline nit (surprised editors don't append it by default)

api/src/prover/hypernova/error.rs
```diff
@@ -0,0 +1,145 @@
...
+            NovaProofError::IOError(e) => IOError(e),
+            NovaProofError::CircuitError(e) => CircuitError(e),
+            NovaProofError::SerError(e) => SerError(e),
+            // The above error conversions allow reusing convienence functions
+            // from the nova implemementation in this crate.
+            //
+            // The remaining errors are thrown by functions that are specific to
+            // a proof system, and so shouldn't be shared across the definitions
+            // even with identical names.
+            _ => unimplemented!(),
```
slumber, line 104: Yeah that was me reusing folding error everywhere in the crate.

I think soon we will need to split nova into smaller crates to allow easy wasm compilation, and this is when we'll improve error handling. Can keep this for the time being.

api/src/prover/hypernova/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+
+pub fn run(opts: &VMOpts, pow: bool) -> Result<Trace, ProofError> {
+    super::nova::run(opts, pow).map_err(ProofError::from)
+}
+
+pub fn init_circuit_trace(trace: Trace) -> Result<SC, ProofError> {
+    super::nova::init_circuit_trace(trace).map_err(ProofError::from)
+}
+
+pub fn prove_seq(pp: &PP, trace: Trace) -> Result<IVCProof, ProofError> {
```
slumber, line 34: test or small example would be a nice-to-have to avoid exposing something that potentially doesn't work

sjudson: Added, though let me know if I misunderstood what you were looking for.

slumber: I meant an example similar to https://github.com/nexus-xyz/nexus-zkvm/blob/main/api/examples/prover_run.rs
maybe more concise just to showcase the usage of hypernova, will serve both as a test and a documentation for those who want to try it.

sjudson: I see, yes, will do.

sjudson: @slumber added

tools/src/command/public_params.rs
```diff
@@ -42,6 +42,7 @@ pub(crate) fn setup_params(args: SetupArgs) -> anyhow::Result<PathBuf> {
     } else {
         match vm_config.prover {
             ProverImpl::Jolt => anyhow::bail!("Jolt doesn't require Nova-setup"),
+            ProverImpl::HyperNova => todo!(),
```
slumber, line 45: Consider returning human readable error like "HyperNova isn't supported by the cli" instead of panicking.

Alternatively, remove it from clap traits so it cannot be possibly parsed.

slumber (approved):

tools/src/command/public_params.rs
sjudson: Done.

PR #202, opened 2024-06-25
Rename to Core

sjudson:
This PR renames `api` -> `core` for the unified interface for exposing virtual machine and proving interfaces. The name `core` is less likely to interfere with other usages and more naturally captures its "black-box" usage by the network crate.


network/src/client.rs
```diff
@@ -13,7 +13,7 @@ use crate::{
 
 pub const LOG_TARGET: &str = "nexus-network::client";
 
-// const URL: &str = "http://35.209.216.211:80/api";
+// const URL: &str = "http://35.209.216.211:80/core";
```
slumber, line 16: I don't think this file needs any changes. Probably find+replace mistake

slumber (approved):
Sharing thoughts, approving as an effort to spare `-api` name.

-----------
`core` is used for a crate that **implements** core functionality. See jolt-core, jsonrpsee-core, rand-core etc. Such crates usually get hidden from users.

Api/core crate, as was pointed out earlier, re-exports other crates functionality. Similarly, it's done by user-facing crate, like [jolt](https://github.com/a16z/jolt/blob/dbda0d43775bbda8f5acdd65fc436fb68c621f65/src/lib.rs#L1), [jsonrpsee](https://github.com/paritytech/jsonrpsee/blob/master/jsonrpsee/src/lib.rs), [rand](https://github.com/rust-random/rand/blob/ca9e1198596ef766714c4c4fdb0070a45334b077/src/lib.rs#L97), all depend on their core-counterparts. 

My point here is that this name doesn't fit either.

Some generic name like `nexus-zkvm` or `nexus-snark-vm` etc. could work (btw worth taking it on crate.io until too late). But from what I understand, api/core isn't supposed to be imported by users? So, going in the loop, it's hard for me to come up with appropriate naming because I never understood its purpose. 🤷‍♂️

network/src/client.rs
sjudson: Yep that was, sorry, will fix

sjudson:
> Such crates usually get hidden from users.

Yes, that's the end goal here, to hide this crate from users behind `tools` and `sdk` and the network interfaces.

> Api/core crate, as was pointed out earlier, re-exports other crates functionality

Another option is to do this directly rather than through re-export, by moving `vm` and `nova` and `jolt` inside of `core`. But I have a preference for re-export because then those crates can be versioned separately, as they may be developed separately.

PR #204, opened 2024-06-27
Add a constraint in split128()

yoichi-nexus:
that relates the variable "{scalar}" and the bits "{scalar}_{i}"s.

### DIFF
--- a/vm/src/circuit/riscv.rs
+++ b/vm/src/circuit/riscv.rs
@@ -937,7 +937,9 @@ fn choose(cs: &mut R1CS, result: &str, bit: &str, left: &str, right: &str) {
 }
 
 fn split128(cs: &mut R1CS, scalar: &str, widths: &[usize]) {
-    let BigInt([a, b, _, _]) = cs.get_var(scalar).into_bigint();
+    let BigInt([a, b, c, d]) = cs.get_var(scalar).into_bigint();
+    assert_eq!(c, 0u64);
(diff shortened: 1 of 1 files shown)


slumber (approved):
Change makes sense to me. A test would be nice-to-have, but honestly circuits are still very flawed.

For reference, see [arkworks bits decomposition](https://docs.rs/ark-r1cs-std/latest/ark_r1cs_std/fields/fp/struct.AllocatedFp.html#impl-ToBitsGadget%3CF%3E-for-AllocatedFp%3CF%3E)

You also want to ensure that this value doesn't overflow field characteristic. Asserting that limbs aren't zero doesn't prevent invalid witness from satisfying the circuit. We had some plans on rewriting circuits with arkworks (should've been done from the very beginning, if you ask me), but it got postponed.

Overall, I'm leaning towards scrapping this implementation in favor of the next gen vm.

Otherwise, good find @yoichi-nexus

sjudson (approved):
Yeah agreed with @slumber that this arithmetization doesn't seem very long for this world and that's a good thing, but we'll take any improvements we can find.

yoichi-nexus:
> You also want to ensure that this value doesn't overflow field characteristic

I think I understand. The bits for `0xffffff....fffff` do still satisfy the constraints for some small `scalar`.

yoichi-nexus:
But here we have just 128 bits. Their powered sum can't get close to the modulo.

PR #205, opened 2024-06-28
rsx_i should be computed based on x_i, but not x'_i

hjlpb:


hjlpb:
Discussed in https://github.com/nexus-xyz/nexus-zkvm/discussions/199

yoichi-nexus (approved):
This looks good. For all indices except `i`, the original register values are kept in `rsx_i`, which get used as final values later.

The original equation also happened to be true on vm executions (because for those registers, the new value and the old value are the same), but that's more like an accident.

yoichi-nexus:
Thank you very much for the find and the fix @hjlpb !

PR #206, opened 2024-06-28
Add constraints to store()

yoichi-nexus:
especially about the relationship between "write_mem{8,16,32}" variables and the corresponding "Y" bits.


sjudson (approved):
It's not entirely clear to me that there is any sort of soundness hole here to be filled: the `write_proof` ultimately maintains the integrity of the memory, and I don't quite see how an invalid choice of `Y`, even if uncaught, could be exploited to maliciously alter any future memory reads.

Nonetheless, semantically this constraint should absolutely hold, so a nice find + add.

PR #207, opened 2024-06-29
Add `pre-commit` for Linting

bhoberman:
`pre-commit` is a useful tool that automatically and universally enforces linting standards before each commit. It also lets us use the exact same automation for linting locally as in CI, which saves headaches.

As a rule, `pre-commit` hooks should avoid compiling code or doing anything that risks a serious slowdown, which is why I omitted `cargo check` and `cargo clippy` from these checks; they make most sense left as-is, as separate CI jobs.

I applied the lints I configured to the existing repo, which has unfortunately mostly generated small edits adding or removing spaces. Fortunately, this type of thing only has to happen once.

To set this up locally, you can run
```bash
pip install pre-commit
cd <nexus repo root>
pre-commit install
```


sjudson (approved):
I have no objection to adding this, but should confirm with @slumber

.pre-commit-config.yaml
```diff
@@ -0,0 +1,19 @@
...
+          - id: end-of-file-fixer
+          - id: trailing-whitespace
+          - id: detect-private-key
+    - repo: https://github.com/yoheimuta/protolint
+      rev: v0.50.2 # Select a release here like v0.44.0
+      hooks:
+          - id: protolint
+    - repo: local
+      hooks:
+          - id: cargo fmt
```
slumber, line 15: this job is present in the main workflow
1. means it's done twice
2. in fact, the main job can fail while this one succeeds, because they're not in sync.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,19 @@
...
+      rev: v4.6.0
+      hooks:
+          - id: check-yaml
+          - id: end-of-file-fixer
+          - id: trailing-whitespace
+          - id: detect-private-key
+    - repo: https://github.com/yoheimuta/protolint
+      rev: v0.50.2 # Select a release here like v0.44.0
+      hooks:
+          - id: protolint
```
slumber, line 12: The workflow says there're no files to check.

slumber (commented):
This workflow runs in parallel with the rest, so it doesn't serve as a shortcut to failure because there's no jobs dependencies, unlike in the main workflow.

> which saves headaches

From my experience, precommit hooks are headaches. Any project I worked on stopped using it eventually. This one is implemented as a workflow, not as a githook, so setting it up with `pip` is required because of non-trivial checks like protobuf files linter which will fail ci on your PR.

Obviously, I _can_ follow this routine, it won't take more than couple minutes in total. But this is something I don't _want_ to think about everytime I'm writing code.

bhoberman:
Closing because this seems unwanted. Can be easily re-opened in the future if we want

PR #210, opened 2024-07-02
nexus-vm: allow limiting executed trace len

slumber:
Small change to allow limiting the number of executed instruction on the VM instance.

Combined with the limit of the ELF file size, this can serve as a spam protection for the cloud.
With the current setup executing a program barely takes any time compared to proving it. So gas metering would be an overkill since the type of instruction shouldn't affect prover time.


sjudson (approved):

sjudson:
Looks great.

PR #211, opened 2024-07-02
rpc: remove server implementation

slumber:
As discussed, moving it out of the monorepo.

Since the interface is fixed, client functionality is still available for future sdk.


sjudson (approved):
We can also look to move the client code into `sdk` at some point.

slumber:
> We can also look to move the client code into `sdk` at some point.

The client code is auto generated by the trait. To use it, import `jsonrpsee::Client` and `RpcClient` traits.

https://github.com/nexus-xyz/nexus-zkvm/blob/cedae2329ed8436d6053c5641046f238c6f0e364/network/rpc/server/src/bin/client.rs#L52-L66

Sdk will need to wrap it into a function call, hiding creation of the client with appropriate configs.

PR #213, opened 2024-07-03
revert store constraints

slumber:
This reverts commit c470d637134797162546321140dbb213dd1761a2.

... as it breaks vm circuits.


sjudson (approved):
Approving @yoichi-nexus, but we should investigate how to robustly add such constraints.

slumber:
cc @yoichi-nexus

yoichi-nexus:
I'll first make sure CI fails for the reverted changes. And then I'll continue investigation on what went wrong.

PR #214, opened 2024-07-05
Add and fix smoke tests on prove and verify

yoichi-nexus:
This MR aims at preventing regression of https://github.com/nexus-xyz/nexus-zkvm/pull/213 , where a newly added constraint broke the examples mentioned in README.md.

This MR
- adds a shell script `smoke.sh` that runs `cargo-nexus nexus {run,prove,verify}`
- adds two lightweight examples in `examples/src/bin` that detect the issue fixed in https://github.com/nexus-xyz/nexus-zkvm/pull/213
- adds the new examples in the CI
- fixes the proving implementation so that the new smoke tests pass

### DIFF
--- a/assets/scripts/smoke.sh
+++ b/assets/scripts/smoke.sh
@@ -25,7 +25,7 @@ if [ ! -f "$1" ]; then
 fi
 
 if [ -e $PROJECT_NAME ]; then
-    echo "Error: Directory 'nexus-project-ci' already exists."
+    echo "Error: Directory '$PROJECT_NAME' already exists."
     exit 1
 fi
 
--- a/assets/scripts/smoke.sh
+++ b/assets/scripts/smoke.sh
@@ -1,0 +1,54 @@
+#! /bin/bash
+
+# This script follows steps written in README.md
+# Using the rust file specified as the unique argument.
+
+# Call at the top directory of the nexus-zkvm project
+# For example:
+# ./assets/scripts/e2e.sh examples/src/bin/fib3.rs
+
+# Every command needs to succeed
+set -e
+
+ORIGINAL_DIR=`pwd`
+PROJECT_NAME="nexus-project-ci"
+
+if [ "$#" -ne 1 ]; then
+    echo "Usage: $0 <file.rs>"
+    exit 1
+fi
+
+if [ ! -f "$1" ]; then
+    echo "Usage: $0 <file.rs>"
+    echo "where <file.rs> is a path to a file."
+    exit 1
+fi
+
+if [ -e $PROJECT_NAME ]; then
+    echo "Error: Directory 'nexus-project-ci' already exists."
+    exit 1
+fi
+
+set -x
+
+cp -n Cargo.toml Cargo.toml.bkp
+
+error_handler() {
+    echo "Error occured in e2e.sh: : ${1}. Exiting."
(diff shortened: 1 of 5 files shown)


assets/scripts/smoke.sh
```diff
@@ -0,0 +1,53 @@
...
+
+cargo build --release --package nexus-tools --bin cargo-nexus
+./target/release/cargo-nexus nexus new $PROJECT_NAME
+cp $1 $PROJECT_NAME/src/main.rs
+cd $PROJECT_NAME
+../target/release/cargo-nexus nexus run
+../target/release/cargo-nexus nexus prove
+../target/release/cargo-nexus nexus verify
+cd $ORIGINAL_DIR
+rm -rf $PROJECT_NAME
```
yoichi-nexus, line 53: the `move -f` line like line 40 is missing.

assets/scripts/smoke.sh
```diff
@@ -0,0 +1,54 @@
...
+fi
+
+if [ ! -f "$1" ]; then
+    echo "Usage: $0 <file.rs>"
+    echo "where <file.rs> is a path to a file."
+    exit 1
+fi
+
+if [ -e $PROJECT_NAME ]; then
+    echo "Error: Directory 'nexus-project-ci' already exists."
```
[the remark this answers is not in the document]
yoichi-nexus: Fixed. Thank you.

assets/scripts/smoke.sh
```diff
@@ -0,0 +1,54 @@
...
+
+cargo build --release --package nexus-tools --bin cargo-nexus
+./target/release/cargo-nexus nexus new $PROJECT_NAME
+cp $1 $PROJECT_NAME/src/main.rs
+cd $PROJECT_NAME
+../target/release/cargo-nexus nexus run
+../target/release/cargo-nexus nexus prove
+../target/release/cargo-nexus nexus verify
+cd $ORIGINAL_DIR
+rm -rf $PROJECT_NAME
```
nexodon, line 53: DRY - could have single cleanup() method when done (whether error or success)

vm/src/trace.rs
```diff
@@ -236,7 +236,7 @@ impl<P: MemoryProof> Iterator for BlockIter<'_, P> {
 
         let s = &self.block.steps[self.index];
         let inst = parse_u32(s.inst).unwrap();
-        let mut w = parse_alt(&self.block.regs, s.inst);
+        let mut w = parse_alt(&self.regs, s.inst);
```
sjudson, line 239: Why this change?

assets/scripts/smoke.sh
yoichi-nexus: true

vm/src/trace.rs
yoichi-nexus: self.block.regs is the initial register file for the folding block. It's immutable for the whole k-step execution. Here we need the register file that changes instruction-to-instruction.

yoichi-nexus: Moreover, without the change, the smoke tests fail.

sjudson (approved):

assets/scripts/smoke.sh
yoichi-nexus: moved those common lines into `cleanup()` function.

PR #215, opened 2024-07-08
Add constraints to store()

yoichi-nexus:
especially about the relationship between "write_mem{8,16,32}" variables and the corresponding "Y" bits.

This PR is very similar to https://github.com/nexus-xyz/nexus-zkvm/pull/206 . This time, I added code to set up the witness values of the new variables. I forgot to do that in #206.


yoichi-nexus:
I realized I need to set up the witness values for the new variables. Trying now.

sjudson (approved):
Makes sense, and this looks good.

PR #217, opened 2024-07-10
add checks on slack bits

yoichi-nexus:
@michel-nexus noticed that the unused bits can be checked to be a specific value, like in RISC-V ISA. This PR implements this idea.


vm/src/rv32/parse.rs
```diff
@@ -121,6 +121,8 @@ fn sop(word: u32) -> Option<SOP> {
 fn aop(word: u32) -> Option<AOP> {
```
michel-nexus, line 121: @yoichi-nexus It seems to me that we need to enforce `opcode(word) = 0b0110011` for all these arithmetic operations and return `None` when this is not the case. It seems to me that we also need to enforce `func7(word) = 0b0100000` for `SUB` and `SRA` and  `func7(word) = 0b0000000` for all the remaining arithmetic operations.

michel-nexus: Actually, line 215 seems to be already enforcing the check `opcode(word) = 0b0110011` so it's not clear to me that you need to check it again in the `aop` function. However, the check for `func7(word) = 0b0100000` for `SUB` and `SRA` and  `func7(word) = 0b0000000` for all the remaining arithmetic operations still seems to be needed.

yoichi-nexus: @michel-nexus Please note that the function `aop()` is called both for `ADD` and `ADDI`. `aop()` needs to look at the opcode again to see if it's about `ADD` or `ADDI`. In case of `ADD`, `aop()` needs to enforce `func7 == 0`. In case of `ADDI`, `aop()` doesn't need to enforce anything on `func7`.

Checks about `SUB`, `SRA`, `AND`, `OR`, `XOR` etc are still coming.

yoichi-nexus: I've applied the comments. Especially I separated `aop()` into two functions each dealing with the `ADD` family and `ADDI` family.

vm/src/rv32/parse.rs
```diff
@@ -119,17 +119,35 @@ fn sop(word: u32) -> Option<SOP> {
...
+        _ => return None,
+    };
+    Some(res)
+}
+
+fn aopi(word: u32) -> Option<AOP> {
+    assert_eq!(opcode(word), OPC_ALUI);
+    let res = match (funct3(word), funct7(word)) {
+        (0b000, _) => ADD,
+        (0b001, 0b0000000) => SLL,
```
sjudson, line 143: We should document here with a comment that `funct7` happens to capture the same set of bits as the `imm[11:6]` chunk that distinguishes the type of intermediate-based shift (`slli` vs. `srli` vs. `srai`).

Otherwise at first glance it looks like we're parsing them wrong since they don't include a `funct7` chunk.

vm/src/rv32/parse.rs
```diff
@@ -162,18 +180,28 @@ mod opcodes {
...
         OPC_LUI => LUI { rd: rd(word), imm: immU(word) },
         OPC_AUIPC => AUIPC { rd: rd(word), imm: immU(word) },
         OPC_JAL => JAL { rd: rd(word), imm: immJ(word) },
-        OPC_JALR => JALR {
-            rd: rd(word),
-            rs1: rs1(word),
-            imm: immI(word),
-        },
+        OPC_JALR => {
+            assert_option(funct3(word) == 0)?;
```
sjudson, line 198: Is there a reason this doesn't panic? We can generally panic (`debug_assert!`) within this circuit construction code: for deployment we only ever construct a single step circuit using it, so it'll be caught in CI testing if anything blows up.

vm/src/circuit/riscv.rs
```diff
@@ -454,24 +473,37 @@ fn parse_J(cs: &mut R1CS, J: u32) {
     cs.mul("J=19", "opcode=19", "f3=0"); // addi
     cs.set_var("J=20", 0); // subi does not exist
     cs.mul("J=21", "opcode=19", "f3=1"); // slli
+    cs.nand("J=21", "f7"); // f7 == 0 if slli
```
sjudson, line 476: Similarly, we should document that this isn't actually `f7`, it's `imm[11:6]`, which coincides with `f7`.

vm/src/circuit/riscv.rs
```diff
@@ -433,6 +451,7 @@ fn parse_J(cs: &mut R1CS, J: u32) {
     cs.set_eq("J=2", "opcode=23"); // auipc
     cs.set_eq("J=3", "opcode=111"); // jal
     cs.set_eq("J=4", "opcode=103"); // jalr
+    cs.nand("f3", "opcode=103");
```
sjudson, line 454: It's not clear to me that this check, while valid, is helpful. We should be explicitly checking that the `f3` chunk corresponds to one of the instructions. 

I feel like this would be a good `debug_assert!`, but not something that should be added to the constraints themselves.

sjudson (changes requested):
Overall this looks great.

vm/src/rv32/parse.rs
yoichi-nexus: Added comments.

vm/src/rv32/parse.rs
yoichi-nexus: `parse_u32` was already using `Option` to catch instruction formatting errors. I just reused the path.

sjudson: Ah I missed that.

vm/src/circuit/riscv.rs
yoichi-nexus: I don't get it. Do you want an alternative prover to convince our verifier about an execution where `jalr` contains a non-zero value in bits 12-14 (inclusive)?

Or, is this constraint implied by some other existing constraints?

vm/src/circuit/riscv.rs
sjudson: I don't see it as having `funct7` per the spec:

Screenshot 2024-07-11 at 11 10 52

Am I looking at the wrong version?

yoichi-nexus: I'm seeing the table. Your comment just made me realize that those bits are now not called funct7.

vm/src/circuit/riscv.rs
sjudson: I'm saying that this check is an over-approximation. The property we want is that `funct3` takes on one of a specific set of values (the left column here), not that `funct3` has a non-zero bit:

Screenshot 2024-07-11 at 11 15 21

vm/src/circuit/riscv.rs
yoichi-nexus: I rephrased the comments.

vm/src/circuit/riscv.rs
yoichi-nexus: > cs.nand("f3", "opcode=103");

This is the check making sure that f3 takes a particular value depending on the opcode.

This check doesn't do anything when "opcode=103" is false. Only when "opcode=103" is non-zero, this check forces "f3" to be zero.

yoichi-nexus: > We should be explicitly checking that the f3 chunk corresponds to one of the instructions.

We're already doing it [here](https://github.com/nexus-xyz/nexus-zkvm/pull/217/files#diff-b8c59a4777a514c97d140db23abfeafad66e6b085f22d50b09cef4fff30db3faR273).

yoichi-nexus: > this check is an over-approximation

Can you name a mal-formed JALR instruction that passes the check?

sjudson: Ugh, okay I misread this check/ the comment on the `nand` function. That makes sense, and you can ignore all of this.

vm/src/rv32/parse.rs
sjudson: These weren't really the comments I was looking for. What was confusing wasn't the `_` matching on the lines that use the full intermediate, what was confusing was the explicit matching for the shifts (since they don't use `funct7` at all). If you can add `(technicallly imm[11:5])` like above that'd be great.

sjudson (approved):
One last minor quibble, but otherwise looks good.

vm/src/circuit/riscv.rs
yoichi-nexus: An alternative formulation would be

    cs.mul("J=4", "opcode=103", "f3=0"); // jalr

This might be weaker or equivalent to the current form in the PR.

yoichi-nexus:
> One last minor quibble, but otherwise looks good.

I did a few more git push'es to give better approximation in comment formulation.

vm/src/rv32/parse.rs
```diff
@@ -119,17 +119,35 @@ fn sop(word: u32) -> Option<SOP> {
...
+        (0b111, 0b0000000) => AND,
+        _ => return None,
+    };
+    Some(res)
+}
+
+fn aopi(word: u32) -> Option<AOP> {
+    assert_eq!(opcode(word), OPC_ALUI);
+    let res = match (funct3(word), funct7(word) /* imm[11:5] */) {
+        (0b000, _ /* imm[11:5] can be any */) => ADD,
```
sjudson, line 142: Wait now I'm confused the other way, this is `funct7` for `ADD` (and all the other non-shifts).

vm/src/rv32/parse.rs
```diff
@@ -119,17 +119,35 @@ fn sop(word: u32) -> Option<SOP> {
...
+        (0b001, 0b0000000 /* shift amount is required to be small */) => SLL,
+        (0b010, _ /* imm[11:5] can be any */) => SLT,
+        (0b011, _ /* imm[11:5] can be any */) => SLTU,
+        (0b100, _ /* imm[11:5] can be any */) => XOR,
+        (0b101, 0b0000000 /* shift amount is required to be small */) => SRL,
```
sjudson, line 147: Also I'm confused by the "small" part, this value doesn't have anything to do with the shift amount, right? It's a fixed value to indicate the type of shift, I think.

vm/src/rv32/parse.rs
yoichi-nexus: This is for ADDI. `OPC_ALUI {ADD ... `.

Maybe we are all becoming fan of flat encoding?

sjudson: Ugh, yep, I got turned around.

vm/src/rv32/parse.rs
yoichi-nexus: I can argue both ways, but I'll just say something else, like, SLLI requires special value here.

yoichi-nexus: done.

PR #219, opened 2024-07-12
Add circuit checking one of "J=?" variables contains one

yoichi-nexus:
I could not find a circuit constraint checking that one of "J=?" variables hold. So I added some checks around "J=?" variables.

* I started using `set_bit()` instead of `set_var()` for creating these variables. This forces their values to be zero or one.
* I added a check that the sum of "J=?" variables equals to one. This makes sure only one of these variables contains one (overflow is impossible).

### DIFF
--- a/vm/src/circuit/riscv.rs
+++ b/vm/src/circuit/riscv.rs
@@ -417,7 +417,7 @@ fn parse_shamt(cs: &mut R1CS, Y: u32) {
 fn parse_J(cs: &mut R1CS, J: u32) {
     cs.set_var("J", J);
     for j in 1..=RV32::MAX_J {
-        cs.set_var(&format!("J={j}"), (j == J).into());
+        cs.set_bit(&format!("J={j}"), j == J);
     }
 
     // used by ALU instructions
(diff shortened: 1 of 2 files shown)


vm/src/circuit/riscv.rs
```diff
@@ -417,7 +417,7 @@ fn parse_shamt(cs: &mut R1CS, Y: u32) {
 fn parse_J(cs: &mut R1CS, J: u32) {
     cs.set_var("J", J);
     for j in 1..=RV32::MAX_J {
-        cs.set_var(&format!("J={j}"), (j == J).into());
```
sjudson, line 420: This var is used elsewhere in the circuit (e.g., https://github.com/nexus-xyz/nexus-zkvm/blob/86e57135d0db805838dd9f42e77a1df63b9084a2/vm/src/circuit/riscv.rs#L589). Maybe we could add a test in `riscv.rs` that `set_var` and `set_bit` behave similarly with various arithmetic operations?

yoichi-nexus: It's mostly self-evident from the implementation
```
    pub fn set_bit(&mut self, name: &str, val: bool) -> usize {
        let j = self.new_var(name);
        self.w[j] = if val { ONE } else { ZERO };
        ...
```
and I can add a test about that.

sjudson: Yeah I'm just worried about an optimization causing a regression.

sjudson (approved):

PR #220, opened 2024-07-12
Fix Memory Limit Example

sjudson:
The memory limit example seems to have gotten jumbled at some point. This makes what's occurring a bit more clear. Setting the memory limit down to `16` (or smaller) will cause the execution to fail over.

### DIFF
--- a/examples/src/bin/memlimit.rs
+++ b/examples/src/bin/memlimit.rs
@@ -7,10 +7,10 @@ use core::hint::black_box;
 use nexus_rt::println;
 
 // use 16mb
-// alternatively, remove memlimit argument or otherwise set less than (slightly more than) 6mb and it will fail
-#[nexus_rt::main(memlimit = 16)]
+// alternatively, remove memlimit argument or otherwise set less than (slightly more than) 16mb and it will fail
+#[nexus_rt::main(memlimit = 18)]
 fn main() {
-    let vec = vec![0 as u32; 1500000];
+    let vec = vec![1 as u8; 0x1000000];
     black_box(vec);
 
     println!("Success!!!");
--- a/runtime/asm.S
+++ b/runtime/asm.S
@@ -18,7 +18,7 @@ _start:
     la gp, __global_pointer$
     .option pop
 
-    la sp, __memory_top - 4
+    la sp, __memory_top - 4 /* start growing the stack (down) from here */
     mv fp, sp
 
     jal ra, _get_stack_size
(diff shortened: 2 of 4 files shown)


slumber:
I'm still confused, assembly seems to set `stack_size`, but the example creates `vec`, whose stack size is fixed (probably 12 bytes on riscv32).

What does this parameter do exactly? We should give it an appropirate name and document it.

yoichi-nexus:
I haven't looked at the assembly, so I am not confused yet. The vector contains 16 mega bytes so there has to be so much memory.

yoichi-nexus (approved):
This seems to be an improvement. Now I can see the idea by comparing 0x1000000 and 16 mega.

sjudson:
> I'm still confused, assembly seems to set `stack_size`, but the example creates `vec`, whose stack size is fixed (probably 12 bytes on riscv32).
> 
> What does this parameter do exactly? We should give it an appropirate name and document it.

Ah I understand the issue @slumber. I'll add significantly more information to the comments to clarify.

Basically, we use `__memory_top` to set where the stack starts. It then grows down (as usual).

When we allocate to the heap, it grows up (as usual, never deallocating). If we request memory being used by the stack be given the heap, then we panic. So functionally, whatever is the "size of the stack" is also the memory limit.

slumber:
I ran it with the limit = 1 and it prints no output, but otherwise process returns 0

sjudson:
> I ran it with the limit = 1 and it prints no output, but otherwise process returns 0

Oh, I think what might be happening is that because the process runs out of memory, we no longer have the space to "clean up". Let me think about how best to handle that...

slumber:
> Basically, we use __memory_top to set where the stack starts. It then grows down (as usual).

>When we allocate to the heap, it grows up (as usual, never deallocating). If we request memory being used by the stack be given the heap, then we panic. So functionally, whatever is the "size of the stack" is also the memory limit.

> Let me think about how best to handle that...

rewrite vm :point_right: :point_left:

sjudson:
> > I ran it with the limit = 1 and it prints no output, but otherwise process returns 0
> 
> Oh, I think what might be happening is that because the process runs out of memory, we no longer have the space to "clean up". Let me think about how best to handle that...

So it should return `1`... https://github.com/nexus-xyz/nexus-zkvm/blob/main/runtime/asm.S#L40

I may merge this PR as is and loop back around on the rest of the improvements to the runtime later.

PR #226, opened 2024-07-16
Update header image for readme as well

sjudson:
Was missed in: https://github.com/nexus-xyz/nexus-zkvm/pull/225


yoichi-nexus (approved):
looks good

PR #229, opened 2024-07-16
Fix Runtime Bugs

sjudson:
Bugs introduced in https://github.com/nexus-xyz/nexus-zkvm/pull/208. @yoichi-nexus do you know why that PR cleared CI/CD?

### DIFF
--- a/runtime/src/ecalls.rs
+++ b/runtime/src/ecalls.rs
@@ -30,7 +30,7 @@ mod riscv32 {
         let inp: u32 = 0;
         let mut out: u32 = 0;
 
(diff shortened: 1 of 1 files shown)


yoichi-nexus (approved):
looks good

yoichi-nexus:
> do you know why that PR cleared CI/CD?

no, my suspicion is it's pulling source from github at some point instead of using the current commit.

sjudson:
> no, my suspicion is it's pulling source from github at some point instead of using the current commit.

That may cause this PR to fail CI/CD as well then...

slumber:
Makes sense to add
```sh
cargo check -p example --target=riscv32i-unknown-none-elf
```
to CI.

PR #231, opened 2024-07-16
speedup CI & check runtime build

slumber:
`build` job was used as a prerequisite for tests, now unnecessarily slows down CI because tests need release build

& check examples on riscv target


yoichi-nexus (approved):
looks good.

sjudson (approved):

Issue #218, opened 2024-07-11
Cover test-machines in CI


slumber:
Follow-up issue for https://github.com/nexus-xyz/nexus-zkvm/issues/203

CI should prove test machines execution similar to the recent smoke test. Something like
```rust
let public_params = gen_vm_pp(16, &()).expect("error generating public parameters");
let machines = ["fib31", "bitop", "shift", "sub", "ldst", /* ... */];
for machine in machines {
    let vm_opts = nvm::VMOpts {
        k: 16,
        machine: Some(machine.into()),
        file: None,
    };
    let trace = trace_vm::<MerkleTrie>(&vm_opts, false, false).unwrap();
    let proof = nexus_core::prover::nova::prove_seq(&public_params, trace)
        .expect("error proving execution");
    proof
        .verify(&public_params, proof.step_num() as _)
        .expect("error verifying execution");
}
```

PR #238, opened 2024-07-17
Add tests proving and verifying the test machines

yoichi-nexus:
Closes https://github.com/nexus-xyz/nexus-zkvm/issues/218

### DIFF
--- a/core/src/prover/hypernova/mod.rs
+++ b/core/src/prover/hypernova/mod.rs
@@ -76,4 +76,25 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn prove_verify_test_machine() -> Result<(), ProofError> {
+        use nexus_vm::{machines::MACHINES, trace_vm};
+        let public_params =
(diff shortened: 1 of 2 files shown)


sjudson (commented):
Can we test Jolt as well? None of these test machines use ECALLs, I don't think.

yoichi-nexus:
> Can we test Jolt as well? None of these test machines use ECALLs, I don't think.

Each of these is worth a follow-up issue, mainly due to the involved chores:
* [jolt](https://github.com/nexus-xyz/nexus-zkvm/issues/239): jolt VM needs a vector of ELFInstructions; the test machines contain u32's
* [ecall](https://github.com/nexus-xyz/nexus-zkvm/issues/240): trace_vm will take private inputs, each test machine should be equipped with a tape of private inputs

sjudson (approved):

slumber (approved):
nice, thanks!

slumber:
> Can we test Jolt as well? None of these test machines use ECALLs, I don't think.

Jolt needs an actual elf for memory checking, so it's not compatible

PR #241, opened 2024-07-17
Test against overflow in alloc_

yoichi-nexus:
I was reading the runtime assembly and realized that the allocation function wasn't checking against overflows. So I added the check.


yoichi-nexus:
I realized I need to update two `.a` files.

sjudson (approved):

PR #242, opened 2024-07-18
Add private input to test machines

### LINKED ISSUE #240: [FEATURE]: add test machine with ecall's; prove and verify them in CI
This is a follow-up issue of https://github.com/nexus-xyz/nexus-zkvm/pull/238#pullrequestreview-2183065055

> None of these test machines use ECALLs, I don't think.

yoichi-nexus:
This PR adds a test machine that reads from the private input. Fixes #240 .

### DIFF
--- a/vm/src/machines.rs
+++ b/vm/src/machines.rs
@@ -6,33 +6,48 @@
 use super::{memory::Memory, rv32::SOP};
 use crate::{NexusVM, Regs};
 
+type TestMachine<'a> = (&'a str, fn() -> Vec<u32>, fn() -> Regs, fn() -> Vec<u8>);
+
 /// An array of test machines, useful for debugging and developemnt.
 #[allow(clippy::type_complexity)]
-pub const MACHINES: &[(&str, fn() -> Vec<u32>, fn() -> Regs)] = &[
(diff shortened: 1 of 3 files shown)


vm/src/machines.rs
```diff
@@ -6,35 +6,56 @@
...
-    ("loop10", || loop_code(10), || loop_result(10)),
-    ("fib31", fib31_code, fib31_result),
-    ("bitop", bitop_code, bitop_result),
-    ("branch", branch_code, branch_result),
-    ("jump", jump_code, jump_result),
-    ("ldst", ldst_code, ldst_result),
-    ("shift", shift_code, shift_result),
-    ("sub", sub_code, sub_result),
+pub const MACHINES: &[TestMachine] = &[
+    ("nop10", || nop_code(10), || nop_result(10), || [].to_vec()),
```
sjudson, line 14: Probably a bit cleaner to do these as `vec![]`.

sjudson (changes requested):
Otherwise looks good

vm/src/machines.rs
yoichi-nexus: When I applied this change, clippy said `|| vec![]` is `Vec::new` so I followed that.

PR #250, opened 2024-07-25
Set opt-level to 3

duc-nx:
Set opt-level to 3

Summary:

Test Plan:


Cargo.toml
```diff
@@ -95,12 +95,13 @@ ark-bls12-381 = { git = "https://github.com/arkworks-rs/curves/", rev = "3fded1f
...
 codegen-units = 1
 lto = true
+opt-level = 3
 
 [profile.release-unoptimized]
 inherits = "release"
 
 [profile.release-unoptimized.package.example]
-opt-level = 0
+opt-level = 3
```
yoichi-nexus, line 104: This profile is called `unoptimized`. It's confusing to have it optimized. I think we should rename the profile or leave the optimization level at zero.

yoichi-nexus (changes requested):

Cargo.toml
sjudson: We should just remove `release-unoptimized` from the codebase and replace it with `release`.

This won't automatically update projects people have already made using `cargo nexus new` or `cargo nexus host`, but is much better for the future.

yoichi-nexus:
This PR contains the changes in #249, which is still a draft. So I'm keeping my blocking review.

duc-nx:
Now the profile in cargo.tom is:

```toml
[profile.release]
codegen-units = 1
lto = true
opt-level = 3

[profile.bench]
inherits = "release"
debug = true
```

yoichi-nexus (dismissed):
Same comments as https://github.com/nexus-xyz/nexus-zkvm/pull/249#pullrequestreview-2202079556 apply.

cli/src/command/prove.rs
```diff
@@ -55,7 +55,7 @@ pub struct LocalProveArgs {
 #[derive(Debug, Args)]
 pub struct CommonProveArgs {
     /// Build artifacts with the specified profile. "release-unoptimized" is default.
-    #[arg(long, default_value = "release-unoptimized")]
+    #[arg(long, default_value = "release")]
```
sjudson, line 58: You also need to modify `cli/src/command/new` to not use the `release-unoptimized` profile in guest projects.

sjudson (changes requested):

cli/src/command/prove.rs
yoichi-nexus: Now they are gone.

yoichi-nexus (approved):
Looks good.

cli/src/command/host.rs
```diff
@@ -52,10 +52,10 @@ fn setup_crate(host_path: PathBuf) -> anyhow::Result<()> {
             "# Generated by cargo-nexus, do not remove!\n",
             "#\n",
             "# This profile is used for generating proofs, as Nexus VM support for compiler optimizations is still under development.\n",
-            "[profile.release-unoptimized]\n",
-            "inherits = \"release\"\n\n",
-            "[profile.release-unoptimized.package.guest]\n",
-            "opt-level = 0\n\n",
+            "[profile.release]\n",
```
sjudson, line 55: This is unnecessary, release uses that opt-level by default. I'd remove from here and below.

sjudson (changes requested):

cli/src/command/host.rs
yoichi-nexus: The same comment applies everywhere else.

duc-nx: fixed in 923b498

duc-nx:
PR #252  test the optimized build of SDK in CI., See log here: 

https://github.com/nexus-xyz/nexus-zkvm/actions/runs/10116238507/job/27978718653?pr=252

cli/src/command/host.rs
sjudson: Sorry if I wasn't clear. There should be nothing about `profile.release` whatsoever. We should leave it undefined, so that the defaults are used.

sjudson (changes requested):

cli/src/command/host.rs
duc-nx: Do you mean to delete the all `[profile.release]` section? 
From 

```toml
[profile.release]
codegen-units = 1
lto = true
opt-level = 3

[profile.bench]
inherits = "release"
debug = true
```

to 
```toml
[profile.bench]
inherits = "release"
debug = true
```

duc-nx: updated in 7cd20ba  
The `opt-level` is deleted, the defaults are used.

sjudson (approved):

PR #253, opened 2024-07-26
Clean up smoke.sh

duc-nx:
Clean up smoke.sh

Since we move smoke.sh output dir to /tmp.

The backup file of Cargo.toml is unneccessary anymore.


yoichi-nexus (approved):
Great, thank you.

PR #256, opened 2024-08-01
development experience improvements

onur-ozkan:
I was hacking around the project and realized there are some minor points that could be improved in terms of development experience.

For more context, I recommend reading the commit descriptions.


sjudson:
Thanks for the PR, but these changes reduce the generality of the devex a little to prioritize certain dev setups, and we'd prefer to keep things basic with the assumption that developers can figure out how to customize things to their own setups. So going to close for now.

PR #259, opened 2024-08-01
v0.2.1 with SDK Guest Runtime Fix

sjudson:
This PR creates a v0.2.1 patch release, that fixes an issue where the SDK installation of the guest runtime was not tied to the same release as the SDK itself. This means that the SDK would pick up on changes to the runtime from the HEAD of `main`, which might not be consistent with the linker scripts that it uses (e.g., the changes in https://github.com/nexus-xyz/nexus-zkvm/pull/249). The changes in this PR set the SDK to install the runtime of the same version.


duc-nx (approved):
Thanks Sam

yoichi-nexus (approved):
Looks good.

PR #260, opened 2024-08-01
Sean/docs metadata 2

smart3ll:


sjudson (approved):
Approving even though not all of the metadata seems to be working quite right, but should solve some other issues in the interim.

PR #265, opened 2024-08-08
Add RISC-V Trace benchmark for Core

duc-nx:
Summary:

Test Plan:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/nexus-xyz/nexus-zkvm/pull/265).
* #270
* #267
* #266
* __->__ #265

PR #266, opened 2024-08-08
Set default `K = 16`

duc-nx:
Summary:

Test Plan:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/nexus-xyz/nexus-zkvm/pull/266).
* #270
* #267
* __->__ #266
* #265

### DIFF
--- a/sdk/src/nova/seq.rs
+++ b/sdk/src/nova/seq.rs
@@ -23,7 +23,7 @@ pub use nexus_core::prover::nova::types::SeqPP as PP;
 use std::marker::PhantomData;
 
 // hard-coded number of vm instructions to pack per recursion step
-const K: usize = 64;
+const K: usize = 16;
 
 /// Errors that occur while proving using Nova.
 #[derive(Debug, Error)]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,7 @@ name: CI
 
 on:
   pull_request:
+    types: [opened, synchronize, reopened, labeled]
   push:
     branches:
       - main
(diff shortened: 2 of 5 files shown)


sdk/src/nova/seq.rs
```diff
@@ -153,8 +153,8 @@ impl Prover for Nova<Local> {
 impl Parameters for PP {
     type Error = Error;
 
-    fn generate_for_testing() -> Result<Self, Self::Error> {
-        Ok(gen_vm_pp(K, &()).map_err(ProofError::from)?)
+    fn generate_for_testing(k: usize) -> Result<Self, Self::Error> {
```
sjudson, line 156: You should also modify the `generate` function calls as well.

sjudson (changes requested):

sdk/src/nova/seq.rs
duc-nx: I'm thinking instead of let user modify the value `k`, which they have no clue, I think we can check the total memory of the system and determine the best value `k` for them. What do you think ? @slumber  @sjudson

duc-nx: done in 0ae46af

slumber: I think it's a bad idea and additional maintenance. CLI defaults to `k=16` which is fine for most users.

Making it configurable like [here](https://docs.nexus.xyz/zkvm/configuring-prover#configuring-k) is sufficient.

sjudson: I agree with @slumber, I think this is overly complicated and puts far too strong assumptions on the availability of the memory. I think we can set `k=16` for now and be done with it.

duc-nx: Thanks @slumber  and @sjudson , I update the default `K` from 64 to 16. in 6a1b337

sjudson (approved):

PR #267, opened 2024-08-08
Add Nova Public Parameter benchmark for SDK

duc-nx:
Summary:

Test Plan:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/nexus-xyz/nexus-zkvm/pull/267).
* #270
* __->__ #267

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,3 +90,13 @@ jobs:
         run: |
             cargo bench --bench riscv_machine
 
+  bench-nova-public-params:
+    runs-on: ubuntu-latest
+    if: contains(github.event.pull_request.labels.*.name, 'benchmark')
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run benchmarks
+        run: |
+            cd nova-benches/
+            cargo bench --bench nova_public_parameter_generation
+
--- a/nova-benches/benches/shared/mod.rs
+++ b/nova-benches/benches/shared/mod.rs
@@ -4,6 +4,7 @@ use ark_relations::r1cs::{ConstraintSystemRef, SynthesisError};
 use nexus_nova::StepCircuit;
 use std::marker::PhantomData;
 
+#[allow(dead_code)]
 pub const NUM_WARMUP_STEPS: usize = 10;
 
 pub struct NonTrivialTestCircuit<F> {
--- a/nova-benches/Cargo.toml
+++ b/nova-benches/Cargo.toml
@@ -28,6 +28,10 @@ name = "nova"
 harness = false
 
 [[bench]]
+name = "nova_public_parameter_generation"
+harness = false
+
+[[bench]]
 name = "hypernova"
 harness = false
 
--- a/nova-benches/benches/nova_public_parameter_generation.rs
+++ b/nova-benches/benches/nova_public_parameter_generation.rs
@@ -1,0 +1,36 @@
+use ark_crypto_primitives::sponge::poseidon::PoseidonSponge;
+use ark_pallas::{Fr as CF, PallasConfig as G1};
+use ark_vesta::VestaConfig as G2;
+use criterion::{criterion_group, criterion_main, Criterion};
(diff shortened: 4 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -70,3 +70,29 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: bnjbvr/cargo-machete@main
+
+  bench-riscv:
```
slumber, line 74: Benches for every small commit is too much, for example we don't run nova benchmarks as a workflow

duc-nx: I run this part in my local and it's fast enough to run on CI. It takes 5 minutes CI, I think it's alright. 
Because I want to have continuous benchmarking flow on CI. 
I can add a label feature, whenever the benchmark label is added, the CI workflow benchmark will be executed.

cothan: added in https://github.com/nexus-xyz/nexus-zkvm/pull/265

duc-nx: added in https://github.com/nexus-xyz/nexus-zkvm/pull/265

sjudson (changes requested):
Benching public parameter generation should live in `nova-benches/`, not `sdk/`.

duc-nx:
I had a hard time moving from `sdk` to `nova-benches`, mainly because of the type `PP` is defined in `sdk`. I would need to duplicate the `PP` type and trait in `nova-benches` similar to `sdk`, this is redundant work. 
Can I let the benchmark run under `sdk`?

sjudson:
You don't need to use the `PP` type from `sdk`, you can use the nova specific parameter types (as is used, e.g., in `nova_benches/` at present).

nova-benches/benches/shared/mod.rs
```diff
@@ -4,6 +4,7 @@ use ark_relations::r1cs::{ConstraintSystemRef, SynthesisError};
 use nexus_nova::StepCircuit;
 use std::marker::PhantomData;
 
+#[allow(dead_code)]
```
sjudson, line 7: This shouldn't be dead code.

duc-nx:
moved to `nova-benches/` in [b7d1403](https://github.com/nexus-xyz/nexus-zkvm/pull/267/commits/b7d140338d31ac71f30cc0ab780ac736412e7159)

nova-benches/benches/shared/mod.rs
duc-nx: Since this is a shared code, it is used in other benches, however I don't use the `NUM_WARMUP_STEPS` in my code, so I turn off the clippy warning in compilation.

sjudson: True I guess it is dead for that individual build... I mean it's all test code so I guess that's fine...

.github/workflows/ci.yml
```diff
@@ -90,3 +90,13 @@ jobs:
         run: |
             cargo bench --bench riscv_machine
 
+  bench-nova-public-params:
```
sjudson, line 93: Why not also benchmark the provers themselves?

nova-benches/benches/shared/mod.rs
duc-nx: coincidently, the value of NUM_WARMUP_STEPS is 10, the minimum I need in criterion sample size is 10, I can plug in this value in my code and get rid of the `allow[dead_code]`, but the variable name doesn't quite describe what the situation.

.github/workflows/ci.yml
duc-nx: Good question. I tried, criterion need 10 samples at least, one prover run on CI takes 4-6 minutes, which lead to at least 40 to 60 minutes to benchmark Prover. 
I have plan to move all CI tasks, including benchmark prover to self-host Github Action in near future.

sjudson (approved):
Would be nice to run all of the Nova related benchmarks but this seems like a reasonable start for now.

PR #268, opened 2024-08-09
Create rust-clippy.yml

duc-nx:
Default rust clippy from official rust with code scanning features from Github.


duc-nx:
#269  add this check already.

PR #269, opened 2024-08-09
Add Github Action cache to improve CI time

duc-nx:


slumber (changes requested):
Note that current CI configuration already caches compile artifacts, from main branch test workflow:
```
... Restoring cache ...
Received 22[229](https://github.com/nexus-xyz/nexus-zkvm/actions/runs/10273767219/job/28429080317#step:3:239)8112 of 509254757 (43.7%), 211.6 MBs/sec
Received 478150656 of 509254757 (93.9%), 227.8 MBs/sec
Cache Size: ~486 MB (509254757 B)
```

It does appear to not work now, might be a misconfiguration issue or something else. Quick search gives https://github.com/actions/cache/blob/main/tips-and-workarounds.md#use-cache-across-feature-branches

Tearing out the whole thing isn't proper solution, as there's no guarantee it will not happen again (especially because two workflows function the same way).

Secondly, the whole CI configuration now looks very cluttered compared to main.

duc-nx:
Hmm, I didn't know we have build cache, perhaps because I only look at build time in CI.  When I made this PR, I notice the build time improve. 
For example: the `test-smoke` and `test-sdk` usually take 8 to 10 minutes, now down to 5 to 7 minutes. 

I separate the clippy, because this way we have code scanning features and can be viewed overtime in Github settings. 
Other than that, I group a few build and check together, so they don't have to query the cache if they run separately. 
Because the CI success is determined at the longest task (which is tesk-sdk and test-smoke), so by combing them together doesn't make the CI worst, IMO.

duc-nx:
Updated the Cache and compare the test time between `main` and this branch: 

- Main [Add Nova Public Parameter benchmark for SDK (#267)](https://github.com/nexus-xyz/nexus-zkvm/actions/runs/10495180481)
- This branch [Add Github Action cache to improve CI time](https://github.com/nexus-xyz/nexus-zkvm/actions/runs/10495190210) 


Build time actually improved. (except `tesk_smoke` and `test_sdk` where we do a fresh build).

.github/workflows/rust-clippy.yml
```diff
@@ -0,0 +1,51 @@
...
+    - cron: '33 20 * * 4'
+
+jobs:
+  rust-clippy-analyze:
+    name: Run rust-clippy analyzing
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      security-events: write
+      actions: read # only required for a private repository by github/codeql-action/upload-sarif to get the Action run status
```
sjudson, line 28: Why is this required then if it's a public repo?

.github/workflows/ci.yml
```diff
@@ -8,62 +8,41 @@ on:
...
       - run: assets/scripts/smoke.sh examples/src/bin/fib3_profiling.rs
 
   test-sdk:
     needs: check-build
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - run: rustup target add riscv32i-unknown-none-elf
+      - name: Install tooling and use Rust cache
+        uses: ./.github/actions/rust-cache
```
yoichi-nexus, line 45: Since `rust-cache` does more than just cache, can you rename it to `rust-setup`? This probably helps us finding things later.

.github/workflows/rust-clippy.yml
yoichi-nexus, line 28: Do we need a line that's required for private repositories only?

.github/workflows/ci.yml
```diff
@@ -8,62 +8,41 @@ on:
...
-        with:
-          toolchain: stable
-          target: riscv32i-unknown-none-elf
-      - run: cargo check --all-features
-      - run: cargo check --all-features --examples
+      - name: Install tooling and use Rust cache
+        uses: ./.github/actions/rust-cache
+      - run: cargo fmt --all --check
       - run: cargo check -p example --target=riscv32i-unknown-none-elf
-      - run: cd nova-benches && cargo check --benches
```
yoichi-nexus, line 32: Why was `cd nova-benches` dropped?

.github/workflows/ci.yml
```diff
@@ -8,62 +8,41 @@ on:
...
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions-rust-lang/setup-rust-toolchain@v1
-        with:
-          toolchain: stable
-      - run: cargo test -r --all-features
+      - name: Install tooling and use Rust cache
+        uses: ./.github/actions/rust-cache
+      - run: cargo test --bins --examples --all-features
```
yoichi-nexus, line 28: Why was the `-r` option dropped?

.github/workflows/rust-clippy.yml
```diff
@@ -0,0 +1,51 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
```
yoichi-nexus, line 4: Why is this sentence here? Depending on the reason, you might want the same sentence in `.github/workflows/ci.yml`.

.github/workflows/rust-clippy.yml
```diff
@@ -0,0 +1,51 @@
...
+name: rust-clippy analyze
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "main" ]
+  schedule:
+    - cron: '33 20 * * 4'
```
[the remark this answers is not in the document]
duc-nx: This is the default clippy script from Github without modification. I don't know why either. 
The outcome of this script is clippy view in Code scanner (which is useful if overtime).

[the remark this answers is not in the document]
yoichi-nexus: This can be meaningful when there is a newer version of `clippy`.

.github/workflows/rust-clippy.yml
duc-nx: leaving this option or remove it won't change the outcome. it's easier to copy paste if we have a private repo.

.github/workflows/rust-clippy.yml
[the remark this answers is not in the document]
sjudson: I think we should cut instead of doing this blindly. I assume it's to break CI if someone does something outside a PR, but we have a protected branch.

.github/workflows/ci.yml
duc-nx: Good idea.

.github/workflows/rust-clippy.yml
sjudson: We generally want to be minimalistic about permissions though

duc-nx: We don't, I assume public repo has the permission on by default. So if we leave it there it's useful if we copy paste the script to our private repo.

.github/workflows/ci.yml
duc-nx: My bad, will add it back asap.

.github/workflows/ci.yml
duc-nx: will add it back asap.

.github/workflows/rust-clippy.yml
duc-nx: This is not our script. We should leave it as is in so in the future we get a fresh new one instead of fixing it.

.github/workflows/ci.yml
duc-nx: fixed in [fc33a1d](https://github.com/nexus-xyz/nexus-zkvm/pull/269/commits/fc33a1dfb94e81140eef538e3cfbe04ff9268c63)

.github/workflows/ci.yml
duc-nx: fixed in [fc33a1d](https://github.com/nexus-xyz/nexus-zkvm/pull/269/commits/fc33a1dfb94e81140eef538e3cfbe04ff9268c63)

.github/workflows/rust-clippy.yml
duc-nx: removed in [fc33a1d](https://github.com/nexus-xyz/nexus-zkvm/pull/269/commits/fc33a1dfb94e81140eef538e3cfbe04ff9268c63)

.github/workflows/ci.yml
duc-nx: fixed in [fc33a1d](https://github.com/nexus-xyz/nexus-zkvm/pull/269/commits/fc33a1dfb94e81140eef538e3cfbe04ff9268c63)

.github/workflows/rust-clippy.yml
duc-nx: removed in [fc33a1d](https://github.com/nexus-xyz/nexus-zkvm/pull/269/commits/fc33a1dfb94e81140eef538e3cfbe04ff9268c63)

.github/workflows/rust-clippy.yml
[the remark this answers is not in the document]
duc-nx: removed in [fc33a1d](https://github.com/nexus-xyz/nexus-zkvm/pull/269/commits/fc33a1dfb94e81140eef538e3cfbe04ff9268c63)

.github/workflows/rust-clippy.yml
sjudson: This is just a _caveat emptor_ that you're using third-party code on GH Actions and if something breaks it's not GHs fault. No harm in keeping it or removing it either way, but might make it easier to track down the provenance.

yoichi-nexus: > This is not our script.

Let's add a link to the source. Also, what's the license of the source?

duc-nx: Idk, it's from the Github action item, and I just click on it and make minor modification. 
I'm pretty sure it's open license, because there are plenty projects use this, although I don't have time to find the exact original license of this source.

yoichi-nexus: > I don't have time to find the exact original license of this source.

That sounds like a blocker.

yoichi-nexus (changes requested):
Third party code without traceable permission of the author: https://github.com/nexus-xyz/nexus-zkvm/pull/269#discussion_r1731841666

PR #270, opened 2024-08-13
Add Continuous Benchmarking to CI

duc-nx:
Summary:

This PR add continous profiling benchmark to CI, so we can track the performanc regression overtime.

Test Plan:


sjudson:
What's the difference between this and the content of #265 and #267? Just that it keeps state about previous runs to capture regressions? How will that be shown? Will it fail, or just be reported in the logs?

sjudson:
Closing as stale.

PR #280, opened 2024-08-21
Re-export postcard in runtime

### LINKED ISSUE #277: Re-export `postcard` from `nexus-runtime` to `nexus-sdk`
### What is the problem?
At the moment, in SDK, the host will have to add postcard in #276 . 
When user modify/update the `postcard` crate, it is not guarantee to work. 

We have a problem with Rust 1.80.1 and `postcard = 1.0.9`, thus, by forcing `postcard` in `nexus-runtime` and re-export in `SDK` a robust solution than the current approach.

evan-schott:
Re-exports `postcard` in `runtime` crate to ensure consistent serialization between host, guest and prover.

Resolves #277.

### DIFF
--- a/cli/src/command/host.rs
+++ b/cli/src/command/host.rs
@@ -116,6 +116,18 @@ fn setup_crate(host_path: PathBuf, rev: Option<String>, tag: Option<String>) -> 
         ],
     )?;
 
+    // add postcard because it is used for (de)serializing from/to the input/output tapes
+    cargo(
+        Some(&guest_path),
+        [
+            "add",
+            "postcard@1.0.10",
+            "-F",
+            "alloc",
+            "--no-default-features",
+        ],
+    )?;
+
     let mut fp2 = fs::OpenOptions::new()
         .append(true)
         .open(guest_path.join("Cargo.toml"))?;
--- a/cli/src/command/host.rs
+++ b/cli/src/command/host.rs
@@ -116,19 +116,6 @@ fn setup_crate(host_path: PathBuf, rev: Option<String>, tag: Option<String>) -> 
         ],
     )?;
 
-    // add postcard because it is used for (de)serializing from/to the input/output tapes
-    // Hardcode postcard version to 1.0.8
-    cargo(
-        Some(&guest_path),
-        [
-            "add",
-            "postcard@=1.0.8",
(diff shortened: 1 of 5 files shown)


cli/src/command/host.rs
```diff
@@ -116,19 +116,6 @@ fn setup_crate(host_path: PathBuf, rev: Option<String>, tag: Option<String>) ->
         ],
     )?;
 
-    // add postcard because it is used for (de)serializing from/to the input/output tapes
```
sjudson, line 119: We want to leave this in for now to not break existing programs. We can remove it eventually.

sjudson (changes requested):

runtime/Cargo.toml
```diff
@@ -11,7 +11,7 @@ categories = { workspace = true }
 
 [dependencies]
 nexus-rt-macros = { path = "macros", version = "0.1.0" }
-postcard = { version = "=1.0.8", features = ["alloc"] }
+postcard = { version = "1.0.10", features = ["alloc"] }
```
duc-nx, line 14: we use postcard for no-std env. Can you also add `default_features=False`?

sdk/Cargo.toml
```diff
@@ -14,7 +14,7 @@ serde.workspace = true
 
 nexus-core = { path = "../core", features = ["prover_nova", "prover_jolt", "prover_hypernova"] }
 nexus-macro = { path = "../macro" }
-postcard = { version = "=1.0.8", features = ["alloc"] }
+postcard = { version = "1.0.10", features = ["alloc"] }
```
duc-nx, line 17: we use postcard for no-std env. Can you also add `default_features=False`?

runtime/src/lib.rs
```diff
@@ -13,3 +13,4 @@ pub use nexus_rt_macros::{main, profile};
 
 mod ecalls;
 pub use ecalls::*;
+pub use postcard::*;
```
sjudson, line 16: It would probably be preferable to just do `pub use postcard`, and have users access through `nexus_rt::postcard` to be clear what's going on here.

sjudson (changes requested):

sjudson (approved):

PR #281, opened 2024-08-27
Add feature-gated partial prove interface

sjudson:
This PR adds a feature-gated partial proof interface that allows generating a proof for a subsequence of a trace.

Intended for use in testing.


slumber (approved):
Add a comment mentioning it is for testing purposes? I was at first confused until I saw PR description, there's no use for partial sequential proof.

PR #283, opened 2024-09-20
dense_mlpoly.rs: Fix manipulation of evaluation vector Z by bound functions

asn-d6:
The bound functions were folding the `Z` vector on itself but were not actually truncating it (even though they were changing `self.len`).

Hence, if you called a bound function and then `evaluate()` it would assert because `evaluate()` checks the size of the `Z` vector.

### DIFF
--- a/spartan/src/dense_mlpoly.rs
+++ b/spartan/src/dense_mlpoly.rs
@@ -221,7 +221,7 @@ impl<F: PrimeField> DensePolynomial<F> {
     for i in 0..n {
       self.Z[i] = self.Z[i] + *r * (self.Z[i + n] - self.Z[i]);
     }
-    self.Z.truncate(n);  // Resize the vector Z to the new length
+    self.Z.truncate(n); // Resize the vector Z to the new length
     self.num_vars -= 1;
     self.len = n;
   }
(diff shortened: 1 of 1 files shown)


sjudson:
@asn-d6 thanks for the PR. We'll take a look internally, but I should note as well that this Spartan code is drawn from [MSR's reference implementation](https://github.com/microsoft/Spartan) so you might want to also send this PR upstream there.

asn-d6:
> @asn-d6 thanks for the PR. We'll take a look internally, but I should note as well that this Spartan code is drawn from [MSR's reference implementation](https://github.com/microsoft/Spartan) so you might want to also send this PR upstream there.
> 
> EDIT: There also appears to be a failing check in the CI.

Good point. Opened a PR upstream here: https://github.com/microsoft/Spartan/pull/73

Also pushed a commit that fixes the CI, I think.

sjudson:
Tentatively, as this appears to be a completeness issue rather than a soundness or zero-knowledge issue, I'm going to hold on it for the upstream PR (which is to an actively maintained repo) to be resolved, as we'd like to retain general parity with the reference implementation.

asn-d6:
Upstream PR got merged for what it's worth!

sjudson (approved):
Approved due to upstream merge.

PR #297, opened 2024-12-09
Fix Cargo lock error

duc-nx:
Summary:

Test Plan:


slumber:
Have you tested jolt?

sjudson:
Following on from @slumber, let's stick with 1.77.0 for the time being: there may be a lot of edge cases to upgrading to 1.80.0, so let's just get the CI working and we can consider the upgrade later.

sjudson (changes requested):
Following on from @slumber, let's stick with 1.77.0 for the time being: there may be a lot of edge cases to upgrading to 1.80.0, so let's just get the CI working and we can consider the upgrade later.

sjudson:
Closing in preference of alternate approach.

PR #306, opened 2024-12-10
Update error.rs

donatik27:
core/src/prover/nova/error.rs
Change: occured → occurred (corrected in five places)


donatik27:
@sjudson

sjudson (approved):

PR #317, opened 2024-12-11
Fix: multiple typos of different importance

crStiv:


sjudson (approved):
We've just merged a new PR request template (https://github.com/nexus-xyz/nexus-zkvm/pull/316) that would exclude these changes, but as this PR was opened before that was merged I'll accept it.

PR #325, opened 2024-12-13
fix:  Improve string formatting for better readability Update network.rs

mdqst:
#### Describe your changes.

In the current implementation, the formatting for strings in `fmt::Display` is done using the following syntax:

```rust
write!(f, "{hostname}{port}")
```

While this syntax works correctly, it can be improved for better readability and clarity. The revised approach uses a more conventional and straightforward method of formatting:

```rust
write!(f, "{}{}", hostname, port)
```

This change doesn't affect functionality but makes the code easier to read and maintain, following a more common formatting style in Rust.

### DIFF
--- a/config/src/network.rs
+++ b/config/src/network.rs
@@ -32,7 +32,7 @@ impl fmt::Display for ClientConfig {
             .map(|port| [":", &port.to_string()].concat())
             .unwrap_or_default();
 
-        write!(f, "{hostname}{port}")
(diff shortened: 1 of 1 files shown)


sjudson:
Thanks for the PR. At them moment the network functionality in this repo has been supplanted by https://github.com/nexus-xyz/network-api and will soon be sunset, so we're not looking to update it at this time.

PR #341, opened 2025-01-09
fix(api.rs) - Improve Enum NexusAPI Variants

cryptoryda:
- Improve Enum NexusAPI Variants.

- Implement Better Error Handling: Add custom error types that make error handling in the `NexusAPI::Error` variant more useful.

- Update the Proof Display implementation for better handling and formatting.

### DIFF
--- a/network/src/api.rs
+++ b/network/src/api.rs
@@ -1,4 +1,5 @@
 use serde::{Deserialize, Serialize};
+use std::fmt::{self, Display};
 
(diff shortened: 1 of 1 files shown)


sjudson:
Thanks for the PR. However, the `network` crate is an old prototype that has mostly been superseded by the code in https://github.com/nexus-xyz/network-api, so we're not looking to take any updates to the code as it'll be sunset soon.

PR #342, opened 2025-01-09
fix(ws.rs) - Ensure that proper async handling is done

cryptoryda:
- Readability: Ensure that the code is clean, especially for error handling and asynchronous operations.

- Async/Sync Context: Ensure that proper async handling is done for both the server and client parts.

- Added custom error types for more detailed error handling.

- Refined `client `function to handle errors in a cleaner way and provide better logging.

### DIFF
--- a/network/src/ws.rs
+++ b/network/src/ws.rs
@@ -1,44 +1,53 @@
 use std::future::Future;
 use std::net::SocketAddr;
-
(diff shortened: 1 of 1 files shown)


sjudson:
Thanks for the PR. However, the `network` crate is an old prototype that has mostly been superseded by the code in https://github.com/nexus-xyz/network-api, so we're not looking to take any updates to the code as it'll be sunset soon.

PR #344, opened 2025-01-18
Update verify.rs

Daulox92:
This change standardizes the UTF-8 encoding string format throughout the codebase. The hyphenated form "utf-8" is the official and more widely used format according to Unicode specifications. The rest of the codebase already uses the hyphenated form "utf-8", so this change brings consistency to the error messages.


sjudson:
Closed per new policy to not accept typo changes. What we have here should be unambiguous and no one should be depending on these error messages in an automated way.

PR #345, opened 2025-01-23
Update rust.yml

Daulox92:
Update actions/checkout from v2 to v4 in Spartan workflow

Changes Made:
spartan/.github/workflows/rust.yml:
- uses: actions/checkout@v2
+ uses: actions/checkout@v4

Why:
- Improves performance and security
- Adds better support for modern Git features
- Aligns with other workflows that already use v4
- Follows GitHub Actions best practices

All other workflows in the repository are already using checkout@v4, making this change consistent across the codebase.

### DIFF
--- a/spartan/.github/workflows/rust.yml
+++ b/spartan/.github/workflows/rust.yml
@@ -10,7 +10,7 @@ jobs:
   build_nightly:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


slumber:
Note that this action is part of forked spartan repo and is never run.

Our changes to the fork are outlined in the readme and we want to keep it commit-by-commit reviewable.

PR #347, opened 2025-02-04
fix: replace unwrap() with proper error handling in PCD node workers

CreeptoGengar:
#### Describe your changes.

Replaces unsafe unwrap() calls with proper error validation:


slumber:
Thank you for the PR, however changes to `/network` crate are not accepted. See https://github.com/nexus-xyz/nexus-zkvm/pull/342#issuecomment-2580510086

PR #349, opened 2025-02-04
fix: enhance progress bar overflow handling

avorylli:
#### Description 

This PR improves the handling of progress bar step counter overflow by replacing the `expect()` call with a safe fallback mechanism and proper warning logging.

#### Changes
- Replaced `expect()` with `match` pattern matching
- Added warning logging for overflow cases
- Implemented safe fallback to 0 when overflow occurs
- Improved error visibility

### DIFF
--- a/cli/progress-bar/src/terminal.rs
+++ b/cli/progress-bar/src/terminal.rs
@@ -129,10 +129,13 @@ impl TerminalContext<'_> {
             return Guard { sender: None };
         };
 
-        self.steps_left = self
-            .steps_left
-            .checked_sub(1)
-            .expect("steps number overflow");
+        self.steps_left = match self.steps_left.checked_sub(1) {
+            Some(n) => n,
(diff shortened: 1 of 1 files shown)


cli/progress-bar/src/terminal.rs
```diff
@@ -129,10 +129,13 @@ impl TerminalContext<'_> {
...
         };
 
-        self.steps_left = self
-            .steps_left
-            .checked_sub(1)
-            .expect("steps number overflow");
+        self.steps_left = match self.steps_left.checked_sub(1) {
+            Some(n) => n,
+            None => {
+                tracing::warn!("Progress bar steps number overflow, resetting to 0");
```
slumber, line 135: Calling `display_step` more than declared `num_steps` times is a logical error that should to panic. Moreover, superconsole renders the whole terminal and this log will be lost.

PR #352, opened 2025-02-08
Fix conversion error

intls:
Before opening your pull request, please respond to the following prompts.

_NB: We DO NOT accept typo fixes. Generally, we do not accept edits to comments (starting with `//`) or minor grammatical and technical edits more generally, but do accept substantive fixes and improvements to the content of documentation comments (`///`) and README files._

### DIFF
--- a/cli/progress-bar/src/terminal.rs
+++ b/cli/progress-bar/src/terminal.rs
@@ -43,7 +43,7 @@ impl TerminalHandle {
 
     pub fn new_enabled() -> Self {
         if superconsole::SuperConsole::compatible() {
-            Self(TerminalHandleInner::new().into())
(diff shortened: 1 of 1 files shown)


cli/progress-bar/src/terminal.rs
```diff
@@ -43,7 +43,7 @@ impl TerminalHandle {
 
     pub fn new_enabled() -> Self {
         if superconsole::SuperConsole::compatible() {
-            Self(TerminalHandleInner::new().into())
```
slumber, line 46: This code is already correct https://doc.rust-lang.org/std/option/enum.Option.html#impl-From%3CT%3E-for-Option%3CT%3E

PR #353, opened 2025-02-08
perf(spartan): Optimize SparsePolynomial evaluate to O(n) complexity

VolodymyrBg:
- Pre-compute 1-r[i] values to avoid repeated subtraction
- Remove separate compute_chi function call overhead
- Use direct iteration and fold for efficient computation
- Improve time complexity from O(n log n) to O(n)


sjudson:

PR #358, opened 2025-02-19
fix: fix incorrect macro usage in criterion_group!

famouswizard:
This resolves a bug in the usage of `criterion_group!`.  

No existing issue was linked, but it fixes a compilation issue with benchmark macros.  

This is a minimal change, fixing a specific problem with macro usage. No need to split.  

#### Describe your changes.  
`criterion_group!` requires a list of functions in `targets`, but explicit type specification (`::<G1Projective, Hyrax<G1Projective>>`) should not be inside the macro. This fix corrects that, ensuring benchmarks compile and run correctly.

### DIFF
--- a/spartan/benches/snark.rs
+++ b/spartan/benches/snark.rs
@@ -140,11 +140,11 @@ fn set_duration() -> Criterion {
   Criterion::default().sample_size(10)
 }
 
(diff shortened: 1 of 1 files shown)


sjudson:

PR #359, opened 2025-02-19
fix: fix incorrect usage of OutputSize::INT in HashBytes

indmind:
This fixes a bug where `OutputSize::INT` was incorrectly used instead of `OutputSize::USIZE`.  

No known open issues, but this corrects a potential type mismatch.  

This is a single, self-contained fix that does not require further splitting.  

#### Describe your changes.  
Replaced `OutputSize::INT` with `OutputSize::USIZE` in `HashBytes`. Since `OutputSize` is already a `typenum` constant (e.g., `U32`), using `ToInt` was unnecessary. `USIZE` ensures proper conversion to `usize` and avoids potential compilation issues.

### DIFF
--- a/network/rpc/common/src/hash.rs
+++ b/network/rpc/common/src/hash.rs
@@ -11,7 +11,7 @@ use sha3::{
 /// Prefixes the program bytes when computing hash.
 pub const PKG_VERSION: &str = env!("CARGO_PKG_VERSION");
 
-type HashBytes = [u8; <Sha3_256Core as OutputSizeUser>::OutputSize::INT];
+type HashBytes = [u8; <Sha3_256Core as OutputSizeUser>::OutputSize::USIZE];
 /// Hash type.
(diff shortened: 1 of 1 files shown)


network/rpc/common/src/hash.rs
```diff
@@ -11,7 +11,7 @@ use sha3::{
 /// Prefixes the program bytes when computing hash.
 pub const PKG_VERSION: &str = env!("CARGO_PKG_VERSION");
 
-type HashBytes = [u8; <Sha3_256Core as OutputSizeUser>::OutputSize::INT];
+type HashBytes = [u8; <Sha3_256Core as OutputSizeUser>::OutputSize::USIZE];
```
slumber, line 14: This is the same associated constant, but specified directly for the type instead of using the trait. The size of the array is checked to be of `usize` type, therefore I don't follow why the usage is "incorrect"

PR #364, opened 2025-02-24
fix: Fixed let-else syntax issue with correct if-let handling

famouswizard:
This resolves a bug related to the use of unstable `let-else` syntax, which causes issues with destructuring references and may be incompatible with older Rust versions.

Yes, this would address compatibility issues caused by the unstable `let-else` syntax in older Rust versions.

This fix addresses a specific syntax issue, and it is best handled in a single change to ensure stability across versions of Rust.

#### Describe your changes.  
I’ve replaced the unstable `let-else` syntax with a standard `if let` statement that correctly handles destructuring references (`&mut`), ensuring compatibility with all Rust versions. This fixes the original issue and eliminates potential instability.

### DIFF
--- a/jolt/src/trace.rs
+++ b/jolt/src/trace.rs
@@ -92,11 +92,12 @@ fn update_row_post_eval<M: Memory>(
     }
     if let Some((lop, store_addr)) = store_addr {
         let new_value = vm.mem.load(lop, store_addr).expect("invalid store").0 as u64;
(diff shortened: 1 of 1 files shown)


slumber:
Thank you for the PR. Note that this syntax **is** stable since Rust 1.65 https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html

PR #370, opened 2025-03-12
chore: fixed wrong `array` import

0xbryer:
Bug.  

No.  

It's a simple fix, no need to split.  

#### Describe your changes.  
Replaced `std::array` with `core::array`—the correct module in Rust.


slumber:
Thank you for the contribution, this is the same module re-exported from the `core` crate https://doc.rust-lang.org/src/std/lib.rs.html#478

PR #381, opened 2025-03-30
Refactor duplicate CPU time calculation

VolodymyrBg:
#### Describe your changes.

Extract duplicated time calculation logic in benchmarks/src/utils.rs into a reusable helper function.

Created a new helper function `calculate_cpu_time` to eliminate code duplication when calculating 
CPU time from rusage structs. This function is now used in the `phase_end` function to calculate
both initial and final CPU times, making the code more maintainable and easier to understand.

This refactoring:
- Reduces code duplication
- Improves readability
- Makes the code more maintainable
- Preserves the exact same functionality

### DIFF
--- a/benchmarks/src/utils.rs
+++ b/benchmarks/src/utils.rs
@@ -178,6 +178,14 @@ pub fn phase_start() -> (std::time::Instant, rusage, rusage) {
     )
 }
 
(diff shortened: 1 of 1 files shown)


VolodymyrBg:

PR #386, opened 2025-04-04
Fix broken profiler link 

Hopium21:
The previous link referenced a non-existent or outdated path (/macro/profiler) which would result in a 404 or lead to confusion. The updated link accurately reflects the current file structure within the repository (/tree/main/sdk/macros/profiler), ensuring clarity and proper redirection for users seeking to include the profiler dependency.


sjudson (approved):

sjudson:
Thanks!

PR #387, opened 2025-04-05
fix: validation for `main` function signature

0xbryer:
#### Describe your changes.

i made a few fixes to the validation logic for the `main` function.

specifically:

- The check for the return type was corrected to properly handle `ReturnType::Default` instead of incorrectly checking for `ReturnType::Type`.
- I fixed the handling of function arguments by using the correct `func.sig.inputs` instead of `fn_sig.inputs`.
- I also updated the `quote!` macro to properly reference `::std::main`, ensuring the correct function path.

### DIFF
--- a/runtime/macros/src/entry.rs
+++ b/runtime/macros/src/entry.rs
@@ -31,7 +31,7 @@ pub fn main(_args: TokenStream, input: TokenStream) -> Result<TokenStream, Error
     }
     if !func.sig.inputs.is_empty() {
         return Err(Error::new_spanned(
-            &fn_sig.inputs,
+            &func.sig.inputs,
             "`main` function arguments must each have an associated input handler",
         ));
     }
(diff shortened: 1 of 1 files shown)


0xbryer:

runtime/macros/src/entry.rs
```diff
@@ -31,7 +31,7 @@ pub fn main(_args: TokenStream, input: TokenStream) -> Result<TokenStream, Error
     }
     if !func.sig.inputs.is_empty() {
         return Err(Error::new_spanned(
-            &fn_sig.inputs,
+            &func.sig.inputs,
```
slumber, line 34: `fn_sig` equals `func.sig`

runtime/macros/src/entry.rs
```diff
@@ -43,7 +43,7 @@ pub fn main(_args: TokenStream, input: TokenStream) -> Result<TokenStream, Error
     }
 
     Ok(quote! {
-        const _: fn() = main;
+        const _: fn() = ::std::main;
```
slumber, line 46: This constant is supposed to check the type of the main function, e.g. against `extern` and presence of return value.

Therefore I don't follow this change

slumber (changes requested):
Thank you for the PR. Feel free to elaborate on your changes and reopen it.

PR #390, opened 2025-04-20
Fix incorrect guest program path

PixelPil0t1:
The old path does not exist in the current repo structure, which may confuse new users. The new path points to the actual guest program used in the examples directory.

If there's a more appropriate or canonical guest path to reference, feel free to suggest it — I’ll be happy to update the PR.


sjudson:
The README is correct -- the path used here is within the host project the guide is walking through creating, not within the code repo.

PR #394, opened 2025-05-10
refactor: replace const LOG_SIZE with log_size() method in FrameworkEvalExt

CreeptoGengar:
This PR addresses the TODO in prover/src/extensions/mod.rs by replacing the constant LOG_SIZE with a method log_size() in the FrameworkEvalExt trait. 

This change allows each component to define its own log size dynamically, improving flexibility and maintainability.


slumber:
Thank you for the contribution. This TODO is addressed in the next release of zkVM.

To add more context, this TODO means that the interface should support variable log size extensions, which is a bigger refactoring. The fix will become public along with keccak permutation support.

PR #398, opened 2025-05-17
Optimize ROM Initialization: Use Iterators Instead of BTreeMap for Memory Byte Access

VolodymyrBg:

This PR resolves a feature request: it optimizes the way ROM initialization data is collected in the emulator by replacing unnecessary intermediate allocations with direct iterators.

The changes are tightly coupled and best resolved together because:
- The new iterator methods (addr_val_bytes_iter) in FixedMemory and UnifiedMemory are required to enable the refactor in LinearEmulator::finalize.

#### Describe your changes.

- Added an addr_val_bytes_iter method to FixedMemory and UnifiedMemory to provide efficient, allocation-free iteration over memory bytes as (address, byte) pairs.
- Refactored LinearEmulator::finalize to use the new iterator for ROM initialization, removing the need to allocate and populate a BTreeMap.

### DIFF
--- a/vm/src/emulator/executor.rs
+++ b/vm/src/emulator/executor.rs
@@ -153,6 +153,7 @@ use nexus_common::{
     cpu::{InstructionExecutor, Registers},
     memory::MemAccessSize,
 };
+use num_traits::FromPrimitive;
 use rangemap::RangeMap;
 use std::{
     cmp::max,
@@ -1161,12 +1162,9 @@ impl Emulator for LinearEmulator {
                 let base_address =
                     self.memory_layout.public_input_start() + i as u32 * WORD_SIZE as u32;
                 let word = word_content.to_le_bytes();
-                word.into_iter()
-                    .enumerate()
-                    .map(move |(j, byte)| MemoryInitializationEntry {
-                        address: base_address + j as u32,
-                        value: byte,
-                    })
+                word.into_iter().enumerate().map(move |(j, byte)| {
+                    MemoryInitializationEntry::new(base_address + j as u32, byte)
+                })
             });
 
         let public_io_loc_iter = self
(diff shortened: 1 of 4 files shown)


sjudson:
Thanks for the PR! We need to review and think about how we feel about this approach.

vm/src/emulator/executor.rs
```diff
@@ -1185,20 +1185,20 @@ impl Emulator for LinearEmulator {
...
+            }
+            Some(uidx) => Box::new(
+                self.memory
+                    .addr_val_bytes_iter(uidx)
+                    .expect("invalid static_rom_image_index")
+                    .map(|(addr, byte)| MemoryInitializationEntry {
+                        address: addr,
+                        value: byte,
+                    }),
+            ) as Box<dyn Iterator<Item = MemoryInitializationEntry>>,
```
duc-nx, line 1200: is there a way we can create an iterator without using `Box`?

vm/src/emulator/executor.rs
```diff
@@ -1185,20 +1186,18 @@ impl Emulator for LinearEmulator {
                     })
             });
         // TODO: avoid creating a BtreeMap and produce an iterator directly
```
sjudson, line 1188: Can you remove as no longer TODO?

vm/src/emulator/executor.rs
```diff
@@ -1233,9 +1232,17 @@ impl Emulator for LinearEmulator {
...
         initial_memory.extend(public_input_iter);
 
-        let tracked_ram_size = self
-            .memory_layout
-            .tracked_ram_size(self.initial_static_ram_image.len_bytes() + rom_initialization.len());
+        let tracked_ram_size = self.memory_layout.tracked_ram_size(
+            self.initial_static_ram_image.len_bytes()
+                + match self.static_rom_image_index {
+                    None => 0,
+                    Some(uidx) => self
```
sjudson, line 1239: Would it be possible to do this count the first time through (https://github.com/nexus-xyz/nexus-zkvm/pull/398/files#diff-ac93bf3e647b4debf9b9fe14b595fb6ca897150133c8edc1fd172d4edfd4a883R1191-R1198) and save it to use here, rather than doing the full iteration again?

vm/src/memory/unified.rs
```diff
@@ -274,6 +275,45 @@ impl UnifiedMemory {
...
+    /// Iterator over addresses and values in the fixed memory, given bytewise, to avoid BTreeMap allocation
+    pub fn addr_val_bytes_iter(
+        &self,
+        uidx: (usize, usize),
+    ) -> Result<Either<std::iter::Empty<(u32, u8)>, Box<dyn Iterator<Item = (u32, u8)> + '_>>, MemoryError> {
+        let (store, idx) = uidx;
+        match FromPrimitive::from_usize(store) {
+            Some(Modes::RW) => {
+                if idx < self.frw_store.len() {
+                    Ok(Either::Right(Box::new(self.frw_store[idx].addr_val_bytes_iter())))
```
sjudson, line 287: Is there a way to avoid using `Box` here?

sjudson (changes requested):
Have a few changes that would be ideal.

sjudson (changes requested):

vm/src/memory/unified.rs
```diff
@@ -274,6 +274,72 @@ impl UnifiedMemory {
         }
     }
 
+    /// Iterator over addresses and values in the fixed memory, given bytewise, to avoid BTreeMap allocation
+    pub fn addr_val_bytes_iter(
+        &self,
+        uidx: (usize, usize),
+    ) -> Result<
```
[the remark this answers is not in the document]
sjudson: Is there a reason you went back to using `Box` rather than explicitly cloning and calling `Into`? My sense is that the latter should be more performant since it can do as much work as possible on the stack.

sjudson:
Tests are now failing @VolodymyrBg

sjudson:
Still failing @VolodymyrBg, I think you might've lost the definition of `FixedMemoryAddrValBytesIter`.

vm/src/memory/unified.rs
```diff
@@ -80,6 +80,16 @@ use std::{
 use super::{
     FixedMemory, LoadOp, MemAccessSize, MemoryProcessor, StoreOp, VariableMemory, NA, RO, RW, WO,
 };
+use crate::memory::fixed::FixedMemoryAddrValBytesIter;
+
+// Helper to cast &FixedMemory<RW> to &FixedMemory<RO>
+fn as_ro(mem: &FixedMemory<RW>) -> &FixedMemory<RO> {
+    unsafe { &*(mem as *const FixedMemory<RW> as *const FixedMemory<RO>) }
```
sjudson, line 87: As cryptographic software we have a very strong aversion to using `unsafe` wherever possible.

Moreover, is it not possible to just do

```
self.frw_store[idx].clone().into<FixedMemory<NA>>()
```

which will implicitly use the `From` implementation in `vm/src/memory/fixed.rs`? The compiler should do a good job of making the cloning/conversion quite fast, and though slower than the recast approach is I think worth doing to avoid any `unsafe` issues.

sjudson (changes requested):
The cast is fast, but I think in cryptographic software I'd prefer an additional clone over it for soundness/robustness/security. It should also still be much more performant than the `BTree` construction it'll be replacing as well.

vm/src/emulator/executor.rs
```diff
@@ -1184,21 +1185,55 @@ impl Emulator for LinearEmulator {
...
-        };
-        let rom_iter = rom_initialization
-            .iter()
-            .map(|(addr, byte)| MemoryInitializationEntry {
-                address: *addr,
-                value: *byte,
-            });
 
+        // Helper function to map (addr, byte) to MemoryInitializationEntry
+        fn map_to_mem_init_entry((addr, byte): (u32, u8)) -> MemoryInitializationEntry {
```
sjudson, line 1190: This should be `MemoryInitializationEntry::new()` per idiomatic Rust.

vm/src/emulator/executor.rs
```diff
@@ -1184,21 +1185,55 @@ impl Emulator for LinearEmulator {
...
+            MemoryInitializationEntry {
+                address: addr,
+                value: byte,
+            }
+        }
+        let mut rom_count = 0;
+        let rom_iter = match self.static_rom_image_index {
+            None => std::iter::empty().collect::<Vec<_>>().into_iter(),
+            Some((store, idx)) => {
+                if store == Modes::RW as usize {
```
sjudson, line 1200: You should be able to do this via pattern matching rather than as an if statement, I think?

vm/src/memory/unified.rs
```diff
@@ -355,6 +355,39 @@ impl UnifiedMemory {
             _ => Err(MemoryError::UndefinedMemoryRegion),
         }
     }
+
```
sjudson, line 358: Are these functions being used anywhere anymore? It seems like no. Either they should be used (rather than making the stores themselves `pub` in the underlying struct), or they should be removed.

sjudson (changes requested):
Overall design is coming together, just some last final nits.

PR #402, opened 2025-05-23
fix Update utils.rs

fatwhorefrommd:
fixes a logic error in the `DurationTracker` where the maximum observed duration was being compared against the **minimum** instead of the **previous maximum**. As a result, `self.max` was effectively reset on every update, rather than preserving the true maximum. This ensures that self.max always reflects the largest duration seen so far.


evan-schott (approved):
Looks good!

fatwhorefrommd:

evan-schott:
recheck

PR #409, opened 2025-06-08
Refactor BEQ chip: Pack neq flags into diffs and store in Helper columns

GarmashAlex:
- Implemented the TODO: packed neq_12_flag and neq_34_flag into diff_12 and diff_34, storing them in the Helper1 column instead of separate columns.
- Updated main trace filling and constraint logic to use diffs for equality checks between valueA and valueB 16-bit parts.
- Adjusted all constraints to work with the new diff-based approach, ensuring field-generic compatibility.

### DIFF
--- a/prover/src/chips/instructions/beq.rs
+++ b/prover/src/chips/instructions/beq.rs
@@ -1,4 +1,4 @@
-use num_traits::One;
+use num_traits::{One, Zero};
 use stwo_prover::{
     constraint_framework::EvalAtRow,
     core::fields::{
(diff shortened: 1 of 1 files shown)


sjudson:
Thanks for the PR @GarmashAlex! As it is, we're currently in the early stages of a refactor of the overall prover design (see https://github.com/nexus-xyz/nexus-zkvm/tree/main/prover2), so we're treating the contents of `prover` as fixed (up to any soundness issues) to focus development on the new direction. But we do appreciate the interest in improving the prover design and hope to accept OSS contributions towards it going forward.

PR #416, opened 2025-06-24
Enforce 16-byte alignment for stack_top in LinearMemoryLayout

GarmashAlex:

This resolves a feature: it enforces the correct 16-byte alignment for the stack_top field in the memory layout, as required by the specification.

No, there are no existing issues related to this change.

This PR is minimal and focused solely on enforcing the correct alignment for stack_top.

#### Describe your changes.

This change updates the alignment check for the stack_top field in LinearMemoryLayout::validate from a 4-byte (word) alignment to a 16-byte alignment, as required by the memory layout specification. The check now uses the assert_aligned_to::<0x10>() method from the Alignable trait. This ensures that the stack top is always properly aligned for architectures or operations that require 16-byte boundaries. No other logic or functionality was modified.

### DIFF
--- a/vm/src/emulator/layout.rs
+++ b/vm/src/emulator/layout.rs
@@ -159,7 +159,7 @@ impl LinearMemoryLayout {
         self.public_output.assert_word_aligned();
         self.heap.assert_word_aligned();
         self.stack_bottom.assert_word_aligned();
-        self.stack_top.assert_word_aligned(); // TODO: this should actually be 0x10-aligned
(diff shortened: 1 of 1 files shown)


sjudson (approved):
This may need to change in the future as we add support for using fixed memory models to skip the first pass tracing (see the discussion in [Sec. 2.2.1](https://specification.nexus.xyz/)), but I think it's a reasonable improvement for the moment.

PR #425, opened 2025-07-10
Use start.elapsed() instead of last_tick.elapsed() in Timer

MozirDmitriy:
Replaced the use of last_tick.elapsed() with start.elapsed() in the Timer component's draw_unchecked method.

This change ensures that the timer always displays the total elapsed time since the timer started, which matches typical user expectations for progress indicators.

Previously, last_tick was not updated anywhere except during initialization and finalize, so its use could be misleading and did not provide any additional value.

Using start.elapsed() makes the code simpler, more transparent, and consistent with how elapsed time is calculated elsewhere in the codebase.

### DIFF
--- a/cli/progress-bar/src/component/timer.rs
+++ b/cli/progress-bar/src/component/timer.rs
@@ -26,7 +26,7 @@ impl Component for Timer<'_> {
         if action.is_finished() {
             return Ok(Lines::new());
         }
-        let elapsed: FmtDuration = self.last_tick.elapsed().into();
+        let elapsed: FmtDuration = self.start.elapsed().into();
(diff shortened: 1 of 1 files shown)


MozirDmitriy:

sjudson:
Thanks for the PR! At the moment the `cli` crate is more of a temporary patch, as we plan in the long-term to incorporate direct zkVM usage with the more general [`nexus-cli`](github.com/nexus-xyz/nexus-cli) project. I'm going to leave this PR open because we might at some point bring the progress-bar over into the project and this may be a useful reference, but for the time being I'm going to leave the crate frozen.

PR #428, opened 2025-07-13
fix: sign extension in PC updates for RISC-V compatibility

Galoretka:
The change was made because the previous logic did not correctly handle negative immediates, resulting in incorrect program counter updates for backward branches and jumps. RISC-V requires sign extension to produce a signed value, so the implementation was updated to match the ISA specification and ensure correct execution of all control flow instructions.

### DIFF
--- a/common/src/cpu/pc.rs
+++ b/common/src/cpu/pc.rs
@@ -11,17 +11,17 @@ impl PC {
 
     // Branch: Add immediate value to PC
     pub fn branch(&mut self, imm: u32) {
-        self.value = self.value.wrapping_add(sign_extension_branch(imm));
+        self.value = self.value.wrapping_add(sign_extension_branch(imm) as i32 as u32);
     }
 
     // Jump and Link: Add immediate value to PC
     pub fn jal(&mut self, imm: u32) {
-        self.value = self.value.wrapping_add(sign_extension_jal(imm));
+        self.value = self.value.wrapping_add(sign_extension_jal(imm) as i32 as u32);
     }
 
     // Jump and Link Register: Set PC to rs1 + imm
     pub fn jalr(&mut self, rs1: u32, imm: u32) {
-        self.value = rs1.wrapping_add(sign_extension_jalr(imm));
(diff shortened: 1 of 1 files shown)


duc-nx:
@Galoretka : The sign extension and wrapping addition in U32 is act as i32 addition. Why does it need to convert to `i32`? Can you write a test case to trigger the error you mentioned? 
> resulting in incorrect program counter updates for backward branches and jumps.

Galoretka:
> @Galoretka : The sign extension and wrapping addition in U32 is act as i32 addition. Why does it need to convert to `i32`? Can you write a test case to trigger the error you mentioned?
> 
> > resulting in incorrect program counter updates for backward branches and jumps.

while the wrapping addition works in most cases, using i32 for sign extension is safer, clearer, and fully compliant with the RISC-V specification. This change helps prevent subtle bugs and makes the codebase more maintainable.

Galoretka:
> I still believe the two approaches are equivalent.
> 
> When converting a signed number from a smaller bit width to a larger bit width, sign extension is necessary regardless. This step is unavoidable.
> 
> How is it different to convert a bit representation to i32, and then again to u32?
> 
> We're discussing how -4 is represented in i32 and u32. It's the same representation, right?
> 
> The original purpose of using the u32 data type was to handle sign extension manually. Let's assume we use i32 as the primary type in the emulator — in that case, we lose control over sign extension and zero extension. That’s why we deliberately chose u32. The entire emulator is designed around this assumption, and the use of wrapping_add is intentional and done with care.
> 
> I'm happy to improve the current design if you can point out a case where it doesn't work. There are many ways to accomplish the same goal, but I think neither your approach nor the current one changes the functional outcome.
> 
> I'm not defending bad code — if something is incorrect or I'm unaware of an issue, I'll fix it ASAP.
> 
> At the moment, I see the PR changes as functionally equivalent to the current implementation.

You're right, this is equivalent. Sorry for wasting your time

duc-nx:
@Galoretka : Nope, no worry. Thank you very much for this discussion.

PR #432, opened 2025-07-20
refactor: fix clippy warnings in common-testing

avorylli:
Address clippy warnings


sjudson:
@avorylli looks like `cargo fmt` is complaining.

PR #440, opened 2025-07-25
Replace panic with error handling for unsupported custom opcodes in instruction encoder

maximevtush:
**Description:**  
This PR improves the robustness of the instruction encoder by replacing the use of `assert_eq!` (which causes a panic) with proper error handling for unsupported custom opcodes.  
- The function `encode_instruction` now returns a `Result`, allowing the caller to gracefully handle unsupported opcodes instead of crashing the application.
- This change increases the safety and reliability of the code, especially when processing untrusted or malformed input.

### DIFF
--- a/common/src/riscv/encoder.rs
+++ b/common/src/riscv/encoder.rs
@@ -136,9 +136,10 @@ pub fn encode_instruction(instruction: &Instruction) -> Result<u32, EncodeError>
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub enum EncodeError {
-    UnsupportedCustomOpcode(u32),
+#[derive(Debug, Error, PartialEq)]
+pub enum OpcodeError {
+    #[error("Cannot convert non-builtin opcode to BuiltinOpcode: {0}")]
+    OpcodeNotBuiltin(Opcode),
 }
 
 #[cfg(test)]


sjudson:
@maximevtush tests are failing

common/src/riscv/encoder.rs
```diff
@@ -118,16 +124,23 @@ pub fn encode_instruction(instruction: &Instruction) -> u32 {
...
+        if instruction.opcode.raw != KECCAKF_OPCODE {
+            return Err(EncodeError::UnsupportedCustomOpcode(
+                instruction.opcode.raw.into(),
+            ));
+        }
+        Ok(encode_s_type(instruction).to_le())
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
```
sjudson, line 139: Why not use the existing `OpcodeError` type (https://github.com/nexus-xyz/nexus-zkvm/blob/main/common/src/error/opcode.rs#L4-L9)?

sjudson (changes requested):

common/src/riscv/encoder.rs
```diff
@@ -118,16 +124,24 @@ pub fn encode_instruction(instruction: &Instruction) -> u32 {
...
+        if instruction.opcode.raw != KECCAKF_OPCODE {
+            return Err(EncodeError::UnsupportedCustomOpcode(
+                instruction.opcode.raw.into(),
+            ));
+        }
+        Ok(encode_s_type(instruction).to_le())
     }
 }
 
+#[derive(Debug, Error, PartialEq)]
```
sjudson, line 139: Since `OpcodeError` already exists I meant to use add `UnsupportedCustomOpcode` to it as a new error within the enum.

sjudson (changes requested):

PR #442, opened 2025-07-26
Syntax error in variable name

savvar9991:
1) Syntax error in variable name: In Rust, variable names cannot contain curly braces {idx}. Using DYNAMIC_PRECOMPILE_{idx} as a variable name will cause a compilation error (expected identifier, found ...). This is a clear error, as the code is not valid Rust.
2)In Rust, the link_section attribute requires a string in quotes. In the source code, the string .note.nexus-precompiles is specified without quotes, which is a syntax error and will lead to a compilation error (expected string literal).


sjudson (approved):

sjudson:
@savvar9991 PR looks good, but you'll need to follow the CLA instructions

PR #443, opened 2025-07-28
fix: correct S-type instruction encoding shift for immediate bits

MozirDmitriy:
Fix incorrect shift value in encode_s_type function for RISC-V S-type instructions.

The immediate value in S-type instructions (like SW, SH, SB) is split into two parts:
- imm[11:5] should be placed in bits [31:25] of the instruction
- imm[4:0] should be placed in bits [11:7] of the instruction

The previous implementation used a shift of 20 for imm_11_5, which placed the bits
in positions [26:20] instead of the correct [31:25]. This has been corrected to
use a shift of 25, ensuring proper RISC-V instruction encoding.

This fix ensures that S-type instructions are encoded according to the RISC-V
specification and will generate correct machine code.


duc-nx:
@MozirDmitriy The test is not passed, can you take a look?

PR #449, opened 2025-08-11
sdk/cargo: avoid passing --target for native builds; fix output path

radik878:
Problem:
Passing `--target native` causes Cargo/rustc to fail because there is no such built-in target.
I reproduced this by running:
  cargo build -p sdk --target native
which errored with: "could not find specification for target 'native'".

Fix:
- For native builds, do not pass `--target`; let Cargo infer the host target.
- Keep `--target riscv32im-unknown-none-elf` only for cross builds.
- Compute artifact path conditionally:
  - native: {dest}/{profile}/{prog}
  - cross:  {dest}/{target}/{profile}/{prog}

Result:
`cargo check -p nexus-sdk` succeeds after the change.

### DIFF
--- a/sdk/src/compile/cargo.rs
+++ b/sdk/src/compile/cargo.rs
@@ -101,6 +101,7 @@ impl Compile for Compiler<CargoPackager> {
         let cargo_bin = std::env::var("CARGO").unwrap_or_else(|_err| "cargo".into());
         let mut cmd = Command::new(cargo_bin);
 
+        // Base args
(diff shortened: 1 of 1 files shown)


radik878:

sjudson (approved):
Thanks for the catch, I think `"native"` was a testing placeholder for inferring the native target, that got missed during review and accidentally merged in permanently. But you're right, just skipping the flag entirely is much cleaner.

PR #451, opened 2025-08-12
 Fix memory indexing bug in MemorySegmentImage

anim001k:
## Description

Fixed logical error in `Index`/`IndexMut` trait implementations for `MemorySegmentImage` that could cause out-of-bounds access and incorrect data retrieval.

### DIFF
--- a/vm/src/memory/memory_image.rs
+++ b/vm/src/memory/memory_image.rs
@@ -177,6 +177,8 @@ impl MemorySegmentImage {
             .enumerate()
             .map(|(i, &v)| (i as u32 * WORD_SIZE as u32 + self.base, v))
     }
+
+
 }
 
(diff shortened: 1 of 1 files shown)


sjudson:
@anim001k that `u32` and `usize` model different notions (index by address and directly index into the underlying data structure) is not existing in the implementation before this PR, and to me is quite unnatural -- I would never expect the difference between those two interfaces to be covered by such a minute typing difference, but rather would expect separate function calls.

How about this. Let's have `Index<u32>` and `Index<usize>` have the functionality you use here for `Index<u32>` (which is what, I think, was intended). You can then introduce `index_direct` functions for either both or for just `usize` which do what you've implemented here for `Index<usize>`.

PR #454, opened 2025-08-14
Preserve original function attributes in pprof macro

Forostovec:
Reason: The macro did not re-emit input_fn.attrs, causing loss of function attributes and doc comments (e.g., #[inline], #[must_use], docs). This could affect inlining, linting, and generated documentation.
Change: Capture input_fn.attrs and expand them before the generated function signature (#(#attrs)* #vis #sig { ... }).
Goal: Preserve all original attributes and documentation while keeping profiling behavior unchanged.


Forostovec:
> @Forostovec also notice the failing `cargo fmt` check

Yes, done. interesting why this happened with whitespace tab/space changes

PR #455, opened 2025-08-19
fix(sdk/stwo): remove unnecessary to_owned() in COBS encoding

Fibonacci747:
Replaced to_owned() with direct reference encoding in COBS paths to avoid adding Clone/ToOwned bounds that are not present in Prover trait signatures and are unnecessary for postcard::to_stdvec_cobs.

Behavior preserved: two-step empty check (no COBS for empty payloads) and 4-byte padding remain unchanged; aligns with existing style where COBS encoding operates on references (see Viewable digest helpers).

Impact: eliminates a compilation hazard with generic S/T (no hidden Clone requirement) and avoids redundant allocations.

tests passed


Fibonacci747:

PR #458, opened 2025-08-25
fix: J-type display immediate and remove unreachable JAL arm

sashass1315:
- Use op_c as the J-type immediate in j_type_to_string to match encoder and prover usage.
- Remove unreachable JAL arm from i_type_to_string to avoid confusion, since J-type formatting is handled by j_type_to_string.
- tests pass


sashass1315:

PR #461, opened 2025-09-01
Adjust ProgramMemCheck lookup tuple size to match actual tuple structure

radik878:
Update LOOKUP_TUPLE_SIZE in ProgramMemCheckChip from 3 * WORD_SIZE to 2 * WORD_SIZE_HALVED + WORD_SIZE, aligning the declared relation size with the actual tuples used (Pc halfwords, instruction halfwords, counter bytes). No other logic changes; existing asserts already matched the new size.

### DIFF
--- a/prover/src/chips/memory_check/program_mem_check.rs
+++ b/prover/src/chips/memory_check/program_mem_check.rs
@@ -27,7 +27,7 @@ use crate::{
 /// ProgMemCheckChip needs to be located after CpuChip
 pub struct ProgramMemCheckChip;
 
-const LOOKUP_TUPLE_SIZE: usize = 3 * WORD_SIZE;
+const LOOKUP_TUPLE_SIZE: usize = 2 * WORD_SIZE_HALVED + WORD_SIZE;
 stwo_constraint_framework::relation!(ProgramCheckLookupElements, LOOKUP_TUPLE_SIZE);
 
 impl MachineChip for ProgramMemCheckChip {


prover/src/chips/memory_check/program_mem_check.rs
```diff
@@ -27,7 +27,7 @@ use crate::{
 /// ProgMemCheckChip needs to be located after CpuChip
 pub struct ProgramMemCheckChip;
 
-const LOOKUP_TUPLE_SIZE: usize = 3 * WORD_SIZE;
+const LOOKUP_TUPLE_SIZE: usize = 2 * WORD_SIZE_HALVED + WORD_SIZE;
```
slumber, line 30: Let's update assertions whenever the final lookup tuple is built to match this constant instead of hardcoded value, e.g. here

https://github.com/nexus-xyz/nexus-zkvm/blob/5b9f1aa48a15bfdc1525b29f421b13cb5bc95be4/prover/src/chips/memory_check/program_mem_check.rs#L254

(and other similar places in this file)

radik878: > Let's update assertions whenever the final lookup tuple is built to match this constant instead of hardcoded value, e.g. here
> 
> https://github.com/nexus-xyz/nexus-zkvm/blob/5b9f1aa48a15bfdc1525b29f421b13cb5bc95be4/prover/src/chips/memory_check/program_mem_check.rs#L254
> 
> (and other similar places in this file)

Yess, done

PR #467, opened 2025-09-08
Refactor bitwise trace to avoid duplicate get_value_b call

phrwlk:
Change execute_step to accept pre-fetched operands (value_b, value_c) and call it from generate_trace_row with locally obtained values. This removes the redundant get_value_b() invocation while keeping behavior identical. ProgramStep::get_value_b/get_value_c are pure and cheap, but eliminating duplicate calls aligns with the style used across other components (e.g., add/sub/shift) and prevents unnecessary work during trace generation.


phrwlk:

slumber (approved):

PR #471, opened 2025-09-17
Rename _next_is_first to _is_last_row in Keccak constraints

MozirDmitriy:
Renamed misleading variable _next_is_first to _is_last_row in:
- prover/src/extensions/keccak/memory_check/constraints.rs
- prover/src/extensions/keccak/round/constraints.rs

Updated TODO comments to reference _is_last_row accordingly.

No logic changes; only naming and comment fixes. Lint passes.

Why this is necessary:
- The referenced preprocessed column IDs (keccak_memory_check_is_last_row, keccak_{index}_is_last_row) semantically indicate “is last row,” not “is first.”
- The old name was a copy/paste artifact used in multiple components and could mislead future contributors and obstruct correct enforcement of the padding transition invariant mentioned in the TODO.
- Aligns naming with actual data semantics across components, improving readability and reducing risk of incorrect future changes.

### DIFF
--- a/prover/src/extensions/keccak/round/constraints.rs
+++ b/prover/src/extensions/keccak/round/constraints.rs
@@ -84,7 +84,7 @@ impl<E: EvalAtRow> KeccakRoundEval<'_, E> {
             // is_padding is the last column in the component trace.
             let [is_padding, _next_is_padding] =
                 self.eval.next_interaction_mask(ORIGINAL_TRACE_IDX, [0, 1]);
-            let _next_is_first = self.eval.get_preprocessed_column(PreProcessedColumnId {
+            let _is_last_row = self.eval.get_preprocessed_column(PreProcessedColumnId {
                 id: format!("keccak_{index}_is_last_row", index = self.index),
             });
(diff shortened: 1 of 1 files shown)


slumber:
>    The referenced preprocessed column IDs (keccak_memory_check_is_last_row, keccak_{index}_is_last_row) semantically indicate “is last row,” not “is first.”

right, "is first" isn't used in the linked code

>   The old name was a copy/paste artifact used in multiple components and could mislead future contributors and obstruct correct enforcement of the padding transition invariant mentioned in the TODO.

This name is intentional, `is_last_row` doesn't imply that the next row is first, the trace must be correctly ordered (coset domain -> circle domain -> bit reverse indices)

>    Aligns naming with actual data semantics across components, improving readability and reducing risk of incorrect future changes.

I believe using both hints on the intended semantics. The todo explains it as well.

To elaborate further: the preprocessed trace must contain unique indices to prohibit prover from omitting calls to `keccak-round` in keccak-f. This is going to be implemented later along with full integration of keccak and register memory component.

PR #474, opened 2025-09-22
fix: handle PRECOMPILE_COUNT symbol in ELF parser

avorylli:
Fix ELF parser to skip PRECOMPILE_COUNT symbol when parsing precompile metadata. 

The parser was failing when encountering non-numeric precompile symbols like PRECOMPILE_COUNT. 

Also correct test expectations to match actual file content.


sjudson:
@avorylli tests are failing

sjudson:
Going to leave this pending since the precompile work is still in an intermediary state, and we may not want to skip this symbol but use it in another way.

PR #475, opened 2025-09-22
fix: resolve path issues with spaces in testing framework

avorylli:
Fix linker script path handling in testing framework to work with directory names containing spaces. 

Copy linker script to temporary directory to avoid path parsing issues during compilation.

### DIFF
--- a/common-testing/src/emulator.rs
+++ b/common-testing/src/emulator.rs
@@ -99,7 +99,29 @@ pub fn parse_output<T: DeserializeOwned>(
 }
 
 /// Create a temporary directory with a new Cargo project that has nexus_rt as a local dependency.
-pub fn setup_guest_project(runtime_path: &PathBuf) -> TempDir {
+fn get_runtime_path() -> PathBuf {
+    // Try to find the runtime directory by looking for Cargo.toml
+    let current_dir = std::env::current_dir().unwrap();
+    let mut path = current_dir.clone();
+    
+    // Look for runtime directory in current and parent directories
+    for _ in 0..5 {
+        let runtime_path = path.join("runtime");
+        if runtime_path.join("Cargo.toml").exists() {
+            return runtime_path;
+        }
+        if let Some(parent) = path.parent() {
+            path = parent.to_path_buf();
+        } else {
+            break;
+        }
+    }
+    
+    // Fallback to original logic
+    current_dir.join("runtime")
+}
+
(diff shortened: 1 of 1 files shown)


sjudson:
@avorylli formatting needs to be fixed

avorylli:
@sjudson  When the project is located in a directory path containing spaces (common on Windows), the linker script path in `RUSTFLAGS` causes compilation failures.

**Example scenario:**
- Project path: `C:\Users\John Doe\Desktop\nexus-zkvm-avon`
- Linker script path: `C:\Users\John Doe\Desktop\nexus-zkvm-avon\runtime\linker-scripts\default.x`

**Before the fix:**
```rust
.env(
    "RUSTFLAGS",
    format!(
        "{compile_flags} -C relocation-model=pic -C panic=abort -C link-arg=-T{}",
        original_linker_script.display()  // Path with spaces breaks linker
    ),
)
```

**Error output:**
`error: could not find linker script: C:\Users\John`

The linker fails to parse the path because it stops at the first space.

**The solution:** Copy the linker script to a temporary directory without spaces, ensuring the path in `RUSTFLAGS` is clean and doesn't contain spaces.

**Reproduction steps:**
1. Clone repository to a path with spaces: `C:\Users\John Doe\Documents\nexus-zkvm`
2. Run integration tests
3. Without fix: compilation fails with "could not find linker script" error
4. With fix: compilation succeeds

common-testing/src/emulator.rs
```diff
@@ -99,7 +99,29 @@ pub fn parse_output<T: DeserializeOwned>(
...
 
 /// Create a temporary directory with a new Cargo project that has nexus_rt as a local dependency.
-pub fn setup_guest_project(runtime_path: &PathBuf) -> TempDir {
+fn get_runtime_path() -> PathBuf {
+    // Try to find the runtime directory by looking for Cargo.toml
+    let current_dir = std::env::current_dir().unwrap();
+    let mut path = current_dir.clone();
+
+    // Look for runtime directory in current and parent directories
+    for _ in 0..5 {
```
sjudson, line 108: Is there a way to make this more "principled" then going an arbitrary depth deep to try and find a valid path? I'm open to trying to provide a measure of "best effort" around this problem, but I'd like to see something a bit more well-formed in terms of an approach.

sjudson (changes requested):

PR #477, opened 2025-09-23
Runtime: replace unreachable_unchecked after SYS_EXIT with safe spin loop

viktorking7:
### Description
- Replaces unsafe `core::hint::unreachable_unchecked()` in the panic handler with a safe infinite spin loop after `SYS_EXIT`.
- Adds a brief comment explaining the rationale.

Why:
- `SYS_EXIT` is expected to terminate, but if it doesn’t (e.g., VM/environment malfunction), hitting `unreachable_unchecked` is UB.
- Spinning avoids UB while preserving the “never return” contract.

Impact:
- Safer runtime behavior in edge cases; no functional changes for normal execution.
- No API changes; no performance impact in the non-failure path.

### DIFF
--- a/runtime/src/runtime.rs
+++ b/runtime/src/runtime.rs
@@ -29,9 +29,11 @@ fn panic(info: &PanicInfo) -> ! {
     let _ = write_output!(0, EXIT_PANIC);
     // Finish with exit syscall.
     let _ = ecall!(SYS_EXIT, EXIT_PANIC);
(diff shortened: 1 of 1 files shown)


sjudson:
Even `unsafe`, the `unreachable_unchecked` is much more semantically useful to the reader than the loop, which is the main role of that code. So I'd prefer to keep it as is.

PR #494, opened 2025-10-10
refactor(sdk): extract duplicate input encoding logic into helper function

jaramyweb3:
## Summary
Extracted duplicate input encoding logic from `run_with_input()` and `prove_with_input()` methods into a reusable `encode_input()` helper function.

## Changes
- Added `encode_input<T: Serialize>()` helper function
- Removed 32 lines of duplicated COBS encoding code
- Both methods now use the shared helper function


sjudson (approved):

jaramyweb3:
@sjudson after running `cargo fmt` locally, the files in the PR don’t get formatted — I don’t understand what the issue is with CI, and also when running `cargo fmt --all --check`, it doesn’t show the files that are in my PR

PR #500, opened 2025-10-23
Optimize ELF file loading by replacing per-byte I/O with bulk read

maradini77:
**Performance improvement** - This optimizes the ELF file loading process for better efficiency.

No existing issues.

This is a minimal, single-purpose change that only affects one function.

#### Describe your changes.

**Problem:** The `ElfFile::from_path()` method was using inefficient per-byte I/O via `Read::bytes()`, which:
- Causes excessive system calls for large ELF files
- Contains a potential panic with `expect("Failed to read byte")`
- Adds unnecessary complexity

**Solution:** Replaced with `std::fs::read()` which:
- Reads the entire file in one efficient operation
- Properly propagates I/O errors without panicking

### DIFF
--- a/vm/src/elf/loader.rs
+++ b/vm/src/elf/loader.rs
@@ -122,7 +122,6 @@ impl ElfFile {
     }
 
     pub fn from_path<P: AsRef<Path> + ?Sized>(path: &P) -> Result<Self, VMError> {
-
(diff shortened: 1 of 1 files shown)


sjudson:
@maradini77 notice failing formatting

sjudson:
@maradini77 tests are failing

PR #504, opened 2025-10-24
refactor(prover/cpu): remove duplicate is_padding declaration in add_constraints

Forostovec:
Removed the second is_padding binding in CpuChip.add_constraints() to avoid variable shadowing and improve readability.
Reused the earlier is_padding value throughout the function; no logic or behavior changes.
first binding - 324 line


slumber:
I think it helps the readability as it's listed along all other op flags -- next to the constraint enforcing that exactly one flag is set.

PR #509, opened 2025-10-28
perf: extract repeated BaseField::from() calls into helper functions

avorylli:
Replace inline BaseField::from() calls with helper functions to avoidf repeated runtime computations in constraint evaluation, improvimg performance and code readability for RISC-V instruction decoding


slumber:
Thank you for the PR.

I'm pretty sure compiler optimizes it already, and the original version reads better. Feel free to provide benchmarks that show improvements.

PR #514, opened 2025-11-01
Fix incorrect operators in LoadStore constraint comments

Evelina331:
Fix the comments incorrectly show `+` operators where the actual implementation uses `-`.


sjudson:
Closing per policy to not accept minor documentary fixes.

PR #518, opened 2025-11-03
Remove unused SYS_OVERWRITE_SP runtime constant

Doryu001:
Drop the unused SYS_OVERWRITE_SP constant from runtime/src/lib.rs eliminate the corresponding #[allow(dead_code)], keeping the syscall list focused on active codes


slumber:
Thank you for the PR,

Please see the doc comment on top of constants definitions. This is a reserved syscall code that we'd want to keep.

PR #519, opened 2025-11-04
Remove redundant helper from simple hash example

Doryu001:
Delete the private `str_to_bytes` wrapper that only forwarded to `as_bytes` inline `large_data.as_bytes()` in the hash example to keep the code leaner


slumber:
Thank you for the PR.

I expect the compiler already inlines it, in either case it doesn't affect functionality.

PR #521, opened 2025-11-05
refactor: replace repetitive calls with macro in load_store

avorylli:
Replace four repetitive `subtract_add_access` calls with a macro to reduce code duplication.

- Add `call_subtract_add_access!` macro
- Use macro for Ram1, Ram2, Ram3, and Ram4 access calls
- Remove TODO comment


slumber:
Thank you for the PR.

I don't think this is worth to be macroed, and the original version reads better.

PR #530, opened 2025-11-14
replace hash_map module import with Entry type import

Doryu001:
This PR refactors the import statement in `syscall.rs` to import only the required `Entry` type instead of the entire `hash_map` module, improving code clarity and consistency.


sjudson:
This is fine as is, thanks for the PR though.

PR #531, opened 2025-11-14
chore(vm): remove unused InternalView import from trace.rs

soniseth0:
Removes unused `InternalView` import from `vm/src/trace.rs`.


sjudson:
@soniseth0 This change is incorrect, that is a necessary import. See the failing tests on the PR for more.

PR #532, opened 2025-11-17
fix: incorrect custom_input example in compile_error!

ANtutov:
The compile_error! message in runtime/macros/src/io.rs incorrectly suggested #[nexus_rt::custom_input(bar)], implying a single-arg attribute, but the custom_input parser requires two arguments: the input identifier or tuple and the handler function identifier. This change updates the example to #[nexus_rt::custom_input((x,y,z), fizz)], aligning the error message with the actual attribute grammar and with existing examples and macro expansion tests in the repository, preventing users from following an invalid example.


ANtutov:

PR #539, opened 2025-11-22
chore(vm): remove unused ELF parser debug helper

0xLogicalx:
eliminate the dead debug_segment_info helper from vm/src/elf/parser.rs, drop the commented-out call so the hot path stays free of unused debugging hooks


sjudson:
Eh, I think this is useful to keep around for debugging.

PR #540, opened 2025-11-23
Fix I/O error handling in ElfFile::from_path

maradini77:
```markdown

Bug fix

No

#### Describe your changes.

Replaces panic-prone file reading in `ElfFile::from_path` with proper error handling.

**Problem:**
The method used `std::io::Read::bytes().expect()` which would panic on any I/O error (corrupted file, permission denied, interrupted read), despite returning `Result<Self, VMError>`.

**Solution:**
Changed to use `fs::read()` which properly converts I/O errors to `ParserError::IOError`, maintaining the error propagation contract and preventing unexpected panics.

**Impact:**
- I/O errors are now properly returned instead of causing crashes
- Maintains consistent error handling throughout the ELF loading pipeline
- Improves robustness when dealing with malformed or inaccessible files
```

### DIFF
--- a/vm/src/elf/loader.rs
+++ b/vm/src/elf/loader.rs
@@ -47,8 +47,6 @@ use crate::{elf::parser, error::VMError, memory::MemorySegmentImage};
 
 use elf::{endian::LittleEndian, ElfBytes};
 use std::fs;
-#[cfg(test)]
-use std::fs::File;
 use std::path::Path;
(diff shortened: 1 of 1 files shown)


sjudson:
@maradini77 tests are failing

sjudson (changes requested):
One minor nit.

PR #552, opened 2025-12-10
fix(macros): correct error messages for input types

PivasDesant:
Fix error messages in handle_input to reflect the actual input type (public/private/custom) instead of always saying "public input". Also correct the comment on line 275 that incorrectly referenced "public output" when it should reference "public/private input".

Error messages now dynamically adapt based on InputType enum, providing accurate feedback to users when using public_input, private_input, or custom_input attributes.


sjudson:
@PivasDesant Not quite. The ones that I said "look correct" do only work over public inputs so the original error messages are correct.

sjudson (approved):

PR #553, opened 2025-12-11
refactor: remove redundant IO entry constructors

Forostovec:
Removed the inherent new constructors from MemoryInitializationEntry and PublicOutputEntry, which duplicated the implementations generated by the io! macro via the IOEntry trait. All internal call sites that used MemoryInitializationEntry::new are updated to use explicit struct literals instead, keeping construction simple and consistent while avoiding two parallel ways to create the same types.


sjudson:
@Forostovec I'm fine with removing the redundant `new()` implementations, but using `::new()` to create the types is quite idiomatic, and I find it generally preferable to direct construction (especially because if in the future you want to add some default information in you don't have to go and alter all the constructor sites). 

So if you'd like to remove those changes and limit this PR to just removing the redundant implementation then I'll be happy to accept it.

PR #556, opened 2025-12-17
perf: reuse pending_logup capacity in LogupTraceBuilder

lisenokdonbassenok:
Use the preallocated capacity of LogupTraceBuilder::pending_logup by extending it instead of replacing the Vec with a newly collected one. This removes an unnecessary allocation on the first call and avoids discarding the existing buffer, while preserving the existing behavior of logup trace generation.


lisenokdonbassenok:

slumber (approved):

PR #559, opened 2025-12-18
fix: clarify handle_input diagnostics for all input types

MoNyAvA:
handle_input now formats error messages using the actual input type (public, private, or custom) when reporting duplicate inputs or inputs missing from the function signature. This avoids misleading diagnostics that previously always mentioned “public input” regardless of the attribute used.

In addition, the error emitted for non-identifier argument patterns has been reworded to state that pattern arguments are not supported and a simple identifier is expected, instead of incorrectly blaming self. These changes only affect user-facing compile-time error messages and do not modify macro expansion semantics.


sjudson:
This duplicates https://github.com/nexus-xyz/nexus-zkvm/pull/552

PR #565, opened 2025-12-21
perf: Remove unnecessary allocations in syscall handlers

0xlupin:
Eliminates redundant clones and allocations in syscall implementation by leveraging Rust's move semantics more effectively.


sjudson (approved):
I went ahead and resolved the issue

PR #576, opened 2026-01-08
perf(vm): Remove unnecessary Step clones in trace tests

0xxFloki:
Removes redundant `clone()` calls in trace module tests by using references instead of owned values where ownership is not required.


sjudson:
This is fine as is for a test suite, especially as these blocks are quite lightweight with small `k`.

PR #577, opened 2026-01-08
Remove unused ELF loader test debug helpers

0xcharry:
Delete unused debug helpers (write_instruction_to_file, write_memory_to_file, debug_elf_file) from vm/src/elf/loader.rs tests, removing dead code and unused imports, keep the existing ELF parsing test unchanged; reduces noise and potential unused warnings


sjudson:
These were deliberately left around as they might be helpful in the near future.

PR #578, opened 2026-01-08
perf(vm): remove redundant clone in unified memory macro

0xxFloki:
Removes unnecessary `.clone()` call in the `add_fixed!` macro within `UnifiedMemory`.


sjudson:
@0xxFloki this is not redundant, see failing tests

PR #579, opened 2026-01-08
perf(vm): optimize public input buffer allocation in HarvardEmulator

0xlupin:
Replaces `concat()` with pre-allocated buffer when building public input with length prefix in `HarvardEmulator::from_elf`.


sjudson:
Rust should be pretty efficient under the hood here with `.concat()`, so going to leave this as is for now.

PR #580, opened 2026-01-08
perf(vm): Remove redundant clone in add_fixed macro

0xlupin:
Eliminates unnecessary `.clone()` call in the `add_fixed!` macro within `UnifiedMemory`.


sjudson:
@0xlupin this is not redundant, see failing tests

PR #585, opened 2026-01-15
Refine precompile metadata error formatting

conomist:
inline format string capture for metadata serialization error message
remove redundant formatting argument in precompile macro generation


slumber:
Thank you for the pr, it's a style choice which doesn't affect functionality

PR #587, opened 2026-01-20
refactor(vm): remove dead code from ELF loader tests

0xlupin:
Removes unused debug helper functions and their imports from the ELF loader test module.


slumber:
Thank you for the PR.

See https://github.com/nexus-xyz/nexus-zkvm/pull/600

PR #592, opened 2026-02-01
refactor(vm): remove unused InternalView import

Aleksandr1732:
Remove unused `InternalView` import from `vm/src/trace.rs`


slumber:
Thank you for the PR.

This import is used as seen in CI.

PR #598, opened 2026-02-12
refactor(runtime/macros): remove redundant clones in macro expansion tests

0xxFloki:
Refactors apply_proc_macro function in macro expansion tests to eliminate unnecessary memory allocations by switching from owned types to borrowed references.


slumber:
Thank you for the PR.

This is fine for tests as it is.

PR #599, opened 2026-02-13
Fix: invalid macro syntax

meelon-dev:

This resolves a documentation bug: incorrect Rust attribute syntax in `runtime/README.md`. Attributes were written without the `#` prefix, which is invalid Rust and would not compile if copied.

No existing issue.

The change is minimal and focused: all affected attribute examples in the `runtime/README.md` are fixed in one commit. Splitting would not provide any additional clarity.

#### Describe your changes.

Replaced incorrect attribute syntax in the README:
- `[nexus_rt::private_input(x)]` > `#[nexus_rt::private_input(x)]`
- `[nexus_rt::custom_input]` > `#[nexus_rt::custom_input]`
- `[nexus_rt::custom_output]` > `#[nexus_rt::custom_output]`

This ensures that documentation examples match valid Rust syntax and prevents compilation errors when copied.


slumber:
Thank you for the PR.

It's clear from the context how macros should be used.

PR #600, opened 2026-02-14
refactor(vm): Remove dead code and replace test suppression with cfg(test)

marukai67:
Cleans up `vm/src/elf/parser.rs` by removing unused code and replacing inappropriate lint suppressions with proper conditional compilation.


slumber:
Thank you for the PR.

The removed comment explains the usecase for the unused core.

PR #608, opened 2026-05-07
fix: replace panicking .expect() / assert! on PC access counter overflow in ProgramMemCheckChip

amathxbt:
## Problem

`ProgramMemCheckChip::fill_main_trace` panics in two places when a PC is accessed more than `u32::MAX` times:

```rust
let new_access_counter = last_access_counter
    .checked_add(1)
    .expect("access counter overflow"); // panic 1
// ...
assert!(!carry_bits[WORD_SIZE - 1]); // panic 2 — redundant overflow check
```

An adversarially crafted program with a tight infinite loop (`j 0`) will eventually push the counter past `u32::MAX`, triggering an unconditional panic that **kills the prover process**. This is a denial-of-service vector: a malicious guest program can crash the prover without producing a verifiable trace.

## Fix

- Replace `.expect()` with a `let…else` that returns early on overflow. The constraint system's carry-overflow constraint (`Don't allow overflow` in `add_constraints`) will reject the resulting trace as unsound — no panic required.
- Downgrade the two redundant `assert!` calls to `debug_assert!` since `checked_add` already guarantees the invariant on the non-overflow path.

### DIFF
--- a/prover/src/chips/memory_check/program_mem_check.rs
+++ b/prover/src/chips/memory_check/program_mem_check.rs
@@ -59,9 +59,14 @@ impl MachineChip for ProgramMemCheckChip {
                 .get(&pc)
                 .unwrap_or(&0u32);
             traces.fill_columns(row_idx, *last_access_counter, Column::ProgCtrPrev);
-            let new_access_counter = last_access_counter
-                .checked_add(1)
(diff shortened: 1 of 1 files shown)


amathxbt:

PR #609, opened 2026-05-14
prover: enforce load/store memory alignment in AIR

### LINKED ISSUE #605: Load/Store AIR Does Not Enforce VM Alignment Rules
# Load/Store AIR Does Not Enforce VM Alignment Rules

## Summary

This issue was reproduced against the official `nexus-zkvm` `0.3.6` release source on the `prover/` STWO implementation.

The VM memory model rejects unaligned halfword and word accesses, but the load/store AIR does not enforce the same rule. A forged trace can therefore describe an unaligned memory access that the VM would reject, yet still produce a proof that verifies.

## Root Cause Analysis

The runtime side treats alignment as part of the VM semantics. The memory layer documents and implements alignment checks for halfword and word accesses under [`vm/README.md`](https://github.com/nexus-xyz/nexus-zkvm/blob/0.3.6/vm/README.md) and the VM memory code.

The proving side only constrains effective-address arithmetic and the byte-level RAM tuples in [`load_store.rs`](https://github.com/nexus-xyz/nexus-zkvm/blob/0.3.6/prover/src/chips/instructions/i/load_store.rs). There is no constraint on the low address bits for `LH`/`LHU`/`SH` or `LW`/`SW`. The relation code even contains an explicit assumption that adding byte offsets without carry is safe because of memory alignment.

The replay PoC starts from an honest aligned `SH`, changes the instruction immediate from `0` to `1`, moves the store record to an odd address, and updates the program word in the view to match. The same test separately confirms that the VM memory model rejects the forged unaligned halfword write, then shows that the proving and verification path still accepts it. The key checkpoints are listed in [attachments/evidence.md](./attachments/evidence.md), and the test output is in [attachments/r06.log](./attachments/r06.log).

The standalone PoC patch is attached as [attachments/r06.patch](./attachments/r06.patch).

## Reproduce Steps

From the extracted source tree:

```bash
cd nexus-zkvm-src
cargo test -p nexus-vm-prover test_unaligned_halfword_store_is_still_provable -- --nocapture
```

Expected result: the test passes. Inside the test, the VM-side write to the forged odd address fails, but the prover and verifier still accept the forged execution trace.

## Impact

This creates a semantic mismatch between the VM and the proof system. The verifier can accept a proof for memory behavior that is invalid under the VM's own execution rules. That is a real soundness problem, not just a runtime bug, because the proof no longer enforces the machine semantics it is supposed to certify.

## Attachments

[attachments.zip](https://github.com/user-attachments/files/27157775/attachments.zip)

Kuhai9801:
## Context

Fixes #605.

The VM rejects unaligned halfword and word memory accesses, but the load/store AIR did not enforce that same rule. That meant a forged trace could describe memory behavior that the VM itself would reject.

This PR makes the load/store AIR enforce the same alignment semantics as the VM for:

- `LH`, `LHU`, `SH` halfword accesses
- `LW`, `SW` word accesses

## What changed

- Added `RamBaseAddrAlignmentQuotient`, a small witness for the low byte of `RamBaseAddr` divided by the required alignment.
- Filled that witness only for halfword and word load/store rows.
- Added `LoadStoreChip` constraints so:
  - halfword rows must satisfy `addr_low = 2 * q`
  - word rows must satisfy `addr_low = 4 * q`
- Reused the existing `Range128Chip` lookup to prove `q` is an actual small integer, not an arbitrary field value.
- Added focused regression tests for unaligned `LH`, `LW`, and `SH`, plus aligned acceptance checks for `LH` and `LW`.

## Review notes

The range check is the key part of the fix. Without it, the equality constraint alone would still allow a forged trace to use a fractional field value for `q`. With `q` constrained to `[0, 127]`, the equations force the low address byte to be correctly aligned.

`RamBaseAddr` limbs are already byte-ranged elsewhere, so enforcing the low byte is enough for these alignment rules.

## Verification

Ran locally with the pinned `nightly-2025-05-09-x86_64-pc-windows-gnu` toolchain:

- `cargo fmt --all --check`
- `cargo test -p nexus-vm-prover aligned -- --nocapture`

The focused test run passed with 4 tests passing and 0 failures. The rejection tests intentionally catch AIR constraint failures from malformed traces, so the panic messages printed during those tests are expected.

### DIFF
--- a/prover/src/chips/instructions/i/load_store.rs
+++ b/prover/src/chips/instructions/i/load_store.rs
@@ -122,9 +122,7 @@ impl MachineChip for LoadStoreChip {
         traces.fill_columns(row_idx, ram_base_address, Column::RamBaseAddr);
 
         let alignment_divisor: Option<u8> = match vm_step.step.instruction.opcode.builtin() {
-            Some(BuiltinOpcode::LH) | Some(BuiltinOpcode::LHU) | Some(BuiltinOpcode::SH) => {
-                Some(2)
-            }
+            Some(BuiltinOpcode::LH) | Some(BuiltinOpcode::LHU) | Some(BuiltinOpcode::SH) => Some(2),
             Some(BuiltinOpcode::LW) | Some(BuiltinOpcode::SW) => Some(4),
             _ => None,
         };
@@ -488,8 +486,7 @@ impl MachineChip for LoadStoreChip {
 
             eval.add_constraint(
                 is_word_access
-                    * (ram_base_addr[0].clone()
-                        - alignment_quotient * BaseField::from(4u32)),
+                    * (ram_base_addr[0].clone() - alignment_quotient * BaseField::from(4u32)),
             );
         }
         let carry_flag = trace_eval!(trace_eval, Column::CarryFlag);
--- a/prover/src/chips/instructions/i/load_store.rs
+++ b/prover/src/chips/instructions/i/load_store.rs
@@ -120,6 +120,24 @@ impl MachineChip for LoadStoreChip {
             add_with_carries(value_a, offset)
         };
         traces.fill_columns(row_idx, ram_base_address, Column::RamBaseAddr);
+
+        let alignment_divisor: Option<u8> = match vm_step.step.instruction.opcode.builtin() {
+            Some(BuiltinOpcode::LH) | Some(BuiltinOpcode::LHU) | Some(BuiltinOpcode::SH) => {
+                Some(2)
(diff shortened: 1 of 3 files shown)


1sgtpepper:

1sgtpepper:

1sgtpepper:
recheck

1sgtpepper:
Hi @sjudson @slumber, could either of you review this when possible? I can’t request reviewers directly on this repo. Thanks!

